Issue #58, opened 2026-05-24
Enable TLS (mTLS) on the agent<->core gRPC channel


neochaotic:
## Why

The AgentService gRPC server currently runs **plaintext** (\`grpc.NewServer()\`
with no transport credentials; \`cmd/leoflow-server/main.go\`). The agent client
already supports TLS 1.2+ (\`internal/agent/dial.go\`) and only sends its token
over a secure channel unless \`allowInsecure\`.

This is the **prerequisite for delivering Variables/Connections to tasks** (#54,
ADR 0021): secrets must not transit a plaintext channel. It also hardens XCom and
state reporting.

## Scope

- Server: configure \`grpc.NewServer\` with \`credentials.NewTLS\` (ideally mTLS so
  the core authenticates the agent's cert in addition to the per-task token).
- Cert management: CA + server cert; for mTLS, a per-agent/client cert (or a
  shared bootstrap cert) distributed to task pods. Config for cert paths.
- Keep an explicit \`allow_insecure\` for local/dev (k3d) only, off by default.
- Until TLS is on, #54's secret delivery must **fail closed** on an insecure
  channel unless explicitly opted in for dev.

Related: #54, ADR 0021, #56.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

neochaotic:
## Certificate management & dev experience

Two distinct TLS surfaces — they don't mix:

### A) Agent↔core gRPC (this issue) — internal, pod→server. Never seen by the browser.
- **Dev / k3d:** runs with **no cert** — use `LEOFLOW_AGENT_ALLOW_INSECURE_SECRETS=true` (plaintext channel, dev only; the secret handlers fail closed otherwise). Zero cert setup to develop/test locally.
- **Production:** server-side TLS, cert via **cert-manager** (auto-renew) or a Helm self-signed cert; the CA (public) distributed to agents via ConfigMap/trust-bundle. **mTLS is defense-in-depth, not required** — the per-task bearer token already authenticates the agent, so TLS just needs to encrypt the channel. For mTLS later: cert-manager-csi-driver issues short-lived per-pod client certs (no Secrets, auto-rotated). If a service mesh (Istio/Linkerd) is present, let it do mTLS transparently.

### B) UI / API (browser-facing) — separate concern, NOT this channel.
- **Dev / local:** UI over plain **HTTP on `http://localhost:8080`** → browser accepts localhost with no warning, **no cert, no import**.
[...]

PR #96, opened 2026-05-26
draft(production): Helm hardening + chart-test CI + deployment guide

neochaotic:
**DRAFT — Production track.** Not for merge yet; this is the staged groundwork for when the Production edition opens. Everything ships **disabled by default** (non-breaking) so it can land safely ahead of time.

### Helm hardening (off by default)
- `templates/hpa.yaml` — `HorizontalPodAutoscaler` (safe with the leader-elected scheduler, ADR 0009).
- `templates/pdb.yaml` — `PodDisruptionBudget`.
- `templates/networkpolicy.yaml` — `NetworkPolicy` (trust boundary: clients → HTTP, Prometheus → metrics, **task pods → gRPC**; permissive egress default to avoid silently breaking DB/Redis/kube-apiserver).
- `templates/servicemonitor.yaml` — Prometheus Operator `ServiceMonitor` for `/metrics` (ADR 0010).
- `values.yaml` — `autoscaling` / `podDisruptionBudget` / `networkPolicy` / `metrics.serviceMonitor`, all `enabled: false`.
- (mTLS already exists via `agentTLS`, #58.)

### CI
- `.github/workflows/chart-test.yaml` — `helm lint` + `kubeconform` now; **kind install + pod-per-task smoke** as the gate to flesh out (the K8s analog of the Lite install gate). Triggers on `helm/**` PRs + manual.

### Docs
- `docs/production-deployment.md` — chart/HA/hardening, signed OCI publishing, layered CI, two-flow GitOps (Argo CD for the platform, image+API for DAGs), Kubeflow/Argo comparison, best practices.
- Pointer + toggles table in `helm/leoflow/README.md`.

## TODO before this becomes a real gate
- SHA-pin the helm/kind/azure actions (ADR 0014).
- Add `.github/ci/kind-datastores.yaml` and complete the pod-per-task smoke (build a sample DAG image, `kind load`, `POST /versions`, trigger a run, assert the task pod `Succeeded`).

## Validated offline
`helm lint` clean · all four new templates render with toggles on · `actionlint` clean · pre-commit gate green.

### DIFF
--- a/.github/workflows/chart-test.yaml
+++ b/.github/workflows/chart-test.yaml
@@ -1,0 +1,104 @@
+# DRAFT (Production track) — Helm chart CI. Mirrors the Lite install gate, but for
+# Kubernetes: lint + schema (every PR touching helm/**), then a real kind cluster
+# install with a pod-per-task smoke. Triggered manually (workflow_dispatch) and on
+# helm/** changes so it stays dormant until the Production front opens.
+#
+# TODO before enabling as a required gate:
+#   - SHA-pin the helm/kind/azure actions (ADR 0014 supply chain).
+#   - Flesh out the pod-per-task smoke (build a tiny DAG image, `kind load`, register
+#     via POST /api/v2/dags/:id/versions, trigger a run, assert the task pod succeeds).
+name: Chart test (DRAFT)
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches: [main]
+    paths: ["helm/**", ".github/workflows/chart-test.yaml"]
+
+permissions:
+  contents: read
+
+jobs:
+  # Fast, always-runnable: chart syntax + rendered-manifest schema.
+  lint:
+    name: helm lint + kubeconform
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+      - uses: azure/setup-helm@v4 # TODO: pin to SHA
+        with:
+          version: v3.16.2
+      - name: helm lint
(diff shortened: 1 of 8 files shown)


neochaotic:
## Fresh-runner contract for this CI smoke (F5 from PR #221 review)

A self-review of PR #221 caught a recurring trap: when a Leoflow contributor dogfoods locally, their machine often has \`pip install leoflow-parser\` from prior work or from the IDE. \`leoflow compile\` then "just works" without \`leoflow setup\` having extracted the parser — because Python finds the module in site-packages.

CI runners do NOT have this state, so the smoke for \`leoflow lite\` (this issue) MUST enforce a fresh-runner contract — otherwise it can pass on every PR while real users hit \`No module named leoflow_parser\` on their first attempt.

**Make explicit in this smoke's runbook:**

- The smoke job runs on a **clean Docker image** (no Leoflow pre-installed, no \`pip install leoflow-parser\`, no pre-existing \`~/.leoflow/\` directory).
- The smoke FAILS if any of these are present at the start of the run.
- Cache buster: a pre-step that verifies \`python3 -c "import leoflow_parser"\` exits non-zero. If it succeeds, the smoke aborts with "fresh-runner contract violated — site-packages has a stale install".
- Same contract applies to the per-distro smoke (busybox-libc / hostile-locale variants).

This is the contract that would have surfaced the BYO + setup ordering gap fixed in PR #221 BEFORE a real user hit it.

References: PR #221 (BYO Python + setup docs), #220 (Pro CI matrix), \`TestCompileFailThenRecoverIsClean\` (proves recovery is clean once the gap is hit).

Issue #98, opened 2026-05-26
Audit unit-test effectiveness: contract-faithful vs coverage-theater


neochaotic:
## Why
Two real regressions shipped despite green unit tests and a 78% coverage floor:
- `POST /clearTaskInstances` → **400** (the UI's real payload doesn't bind to our `clearRequest`).
- `GET /ui/grid/ti_summaries` → **500** on a freshly-created run.

Both endpoints **had tests that passed** — but the tests assert payloads/states **we constructed** (e.g. `task_ids:["extract"]` as plain strings, fixtures with hand-set times), not the **real Airflow contract** / real runtime states. Green coverage, zero confidence. Coverage measured lines executed, not behavior-vs-reality.

## Goal
Audit the unit suite for **contract faithfulness** and shift the testing philosophy:
- **Coverage = floor/hygiene, not a goal.** Stop writing tests *to* the number.
- **Anchor tests to external truth:** real Airflow contract fixtures (captured payloads / the `_private_ui.yaml` & openapi schemas), faithful e2e (real flows against a real server), and **reproduce-the-bug-first** (red before green).
- For the Airflow-compat API surface: **contract tests using the real request/response shapes**, not our own DTOs.

## Concrete tasks
- [ ] Capture/record the **real Airflow request payloads** for the write endpoints (clear, trigger, patch) into `docs/reference/airflow-real/` and assert against them.
- [ ] Rewrite `clearTaskInstances` and `ti_summaries` tests as **contract/e2e** (tracked separately — in progress).
- [ ] Add the `e2e-lite-full` job (real run → clear → logs → assert) as a gate.
- [ ] Sweep `internal/api/*_test.go` for self-consistent fixtures that can't fail against reality; flag/rewrite.
- [ ] Document the policy (coverage floor; tests must be able to fail when reality diverges).

Surfaced during pre-alpha Lite testing on real hardware.

neochaotic:
Deferring to product evolution — closing to keep the backlog actionable (pulling features/fixes on demand rather than tracking a large speculative list). The finding/design is preserved here and reopenable when the work is pulled. Bugs & Pro-alpha gates are kept open in the `pro-alpha` milestone.

PR #99, opened 2026-05-26
fix(api): clear tuple task_ids + client-cancel 499 + 4xx/5xx error logging

neochaotic:
Two real API regressions surfaced by testing Lite on real hardware, plus the observability gap that made them invisible. **All diagnoses anchored to the real Airflow contract / real behavior — not self-made fixtures** (see #98).

## Fixes
- **`clearTaskInstances` 400 → fixed.** The Airflow 3.2 UI sends `task_ids` as `[task_id, map_index]` **tuples** (`[["hello", -1]]`, captured from the SPA bundle), but `clearRequest.TaskIDs` was `[]string` → the real payload never bound → clear 400'd on every click. Now accepts both string and tuple forms (+ `only_running`). **Contract test posts the real tuple shape** — the prior test used our own `["extract"]` and never caught it.
- **`ti_summaries` 500 (the "eternal running" grid) → fixed.** `handleRepoError` mapped every non-NotFound error to 500, including `context.Canceled` — which just means the client aborted (the UI supersedes in-flight grid requests under rapid refresh). The run actually finalizes `success` in the DB; the 500s broke the grid's refresh, so it showed stale "running". Now maps client-cancel to **499**.
- **4xx/5xx error logging.** `AbortProblem` records its detail; `StructuredLogger` logs 5xx at ERROR / 4xx at WARN with the cause. Failing requests were logged only as a bare status code — this is what cracked both bugs and guards future regressions.

## Not bugs (clarified during diagnosis)
- IDE button: present (floating bottom-right), not in the sidebar.
- Empty task log: the starter `hello` task only `return`s; log capture works (verified a `print()` task ships stdout + the Airflow SDK logger).

## Tests
- `TestClearAcceptsAirflowTupleTaskIDs` (contract: real tuple payload).
- `TestHandleRepoErrorClientCancel` (cancel/deadline → 499; NotFound → 404; real error → 500).
- `TestStructuredLoggerSurfacesErrorDetail` (5xx=ERROR, 4xx=WARN, +detail).

## Validation
`go test ./...` exit 0 · `golangci-lint` 0 issues · validated end-to-end on a VM: clear with tuples → 200 (was 400), task log captures stdout + SDK logs, run finalizes success.

Follow-ups: the pre-login flash (client-side, cosmetic) and the test-effectiveness audit (#98).

### DIFF
--- a/internal/api/middleware.go
+++ b/internal/api/middleware.go
@@ -16,7 +16,11 @@ import (
 const (
 	contextKeyUser      = "leoflow.user"
 	contextKeyRequestID = "leoflow.request_id"
-	headerRequestID     = "X-Request-Id"
+	// contextKeyProblemDetail carries the AbortProblem detail so StructuredLogger
+	// can surface WHY a 4xx/5xx happened — otherwise a failing request is logged
+	// only as a status code, with no cause (an observability blind spot).
+	contextKeyProblemDetail = "leoflow.problem_detail"
+	headerRequestID         = "X-Request-Id"
 )
 
 // RequestID assigns a request id (honoring an inbound X-Request-Id) and echoes it.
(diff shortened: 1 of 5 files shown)


neochaotic:
Adding scope: **task_instance_history** table (introduced 2026-06-01 to close #241) grows 1 row per Clear/retry, unbounded today. Should be included in the GC policy — recommended approach: keep last N attempts per TI (e.g., N=10) and/or drop entries where the parent dag_run is older than the run-history retention window. Already FK'd ON DELETE CASCADE from task_instances so deleting a dag_run cleans up its TIs' history automatically; only orphaned-runs cleanup needs awareness.

Issue #88, opened 2026-05-25
Lite: embedded Monaco web editor (the Lite IDE)


neochaotic:
Implements ADR 0025 — the last Lite-exclusive feature: a built-in, minimal web
editor served by the control plane, opened from the Leoflow UI, to edit DAG
projects in the workspace.

**Why Monaco (not code-server/Theia):** full VS-Code-in-browser servers are heavy
(Node, ~150 MB) and break over plain HTTP off-localhost (Service Workers /
Clipboard API / Secure cookies are gated to a secure context). Lite targets an
internal-network/VPN `http://<LAN-IP>`, where they'd need TLS. Monaco is
client-side JS (no Service Worker), served same-origin over HTTP, opened in a new
tab — it passes the rule and stays light. See ADR 0025.

### Plan
1. **Files API** (Go, control plane, Lite-only, workspace-sandboxed):
   `GET /api/v2/ide/tree`, `GET/PUT/DELETE /api/v2/ide/file`, `POST /api/v2/ide/file` (create).
   Strict path confinement to the workspace root (reject abs paths + `..`). Tests.
2. **Editor page**: serve `/ide` — vendored Monaco + a file-tree panel (open / edit
   / save / create / delete), `?theme=light|dark` (`vs`/`vs-dark`).
3. **UI entry**: inject an "IDE" button into the served shell (like the LITE badge),
   opening `/ide` in a new tab; Lite edition only.
4. **Wire workspace**: pass the workspace root to the control plane (config/env).
5. **Docs**: a user-facing "Web editor" section under the Lite docs.

### Notes
- Orthogonal to the executor — edits workspace source on the host; identical under
  `subprocess` and `k8s`.
- Not a full IDE (no extensions/terminal/debug) — by design.
- Security: the files API is a local-disk surface — sandbox strictly; Lite-only.

neochaotic:
Deferring to product evolution — closing to keep the backlog actionable (pulling features/fixes on demand rather than tracking a large speculative list). The finding/design is preserved here and reopenable when the work is pulled. Bugs & Pro-alpha gates are kept open in the `pro-alpha` milestone.

PR #106, opened 2026-05-27
fix(lite): per-user datastore volume + uninstall resets it (keep workspace)

neochaotic:
Completes the root/non-root isolation started in #105 (reconcile-admin handled the *login* symptom; this isolates the *data* and makes a reinstall clean). Validated on a Lima VM as a non-root user.

## #88 — per-user Docker volume
The compose project is now `COMPOSE_PROJECT_NAME=leoflow-$(id -u)`, so root and an unprivileged user get **separate** Postgres/Redis volumes instead of silently sharing one (the data half of the root-vs-non-root collision). Set identically on `up` (lite) and on uninstall's teardown.
- *Validated:* a non-root run creates `leoflow-<uid>_leoflow_pgdata` / `_redisdata`.

## #90 — uninstall resets the datastore, keeps the workspace
`leoflow uninstall` now drops **this user's datastore volumes by default** (so a reinstall starts fresh instead of inheriting a stale admin / old runs), while **keeping the DAG workspace** — only `--purge` removes that. The teardown targets the project **by name** and removes its volumes directly, so it works regardless of which compose file Lite used (a checkout's `docker-compose.dev.yaml`, the managed `~/.leoflow` one, or `--compose`). *(My first cut keyed off a fixed `~/.leoflow/docker-compose.yaml` path and silently left volumes behind when Lite had used a different file — caught and fixed in validation.)*
- *Validated:* uninstall removes the `leoflow-<uid>` volumes and the workspace `dag.py` survives.

Host ports stay shared (5432/6379), so one Lite instance runs at a time — fine for a local single-user tool. Coverage above floor; lint clean.

### DIFF
--- a/internal/cli/dev.go
+++ b/internal/cli/dev.go
@@ -620,6 +620,7 @@ func devComposeUp(ctx context.Context, cmd *cobra.Command, o devOptions) error {
 	devPrintln(cmd.OutOrStdout(), "▸ starting dependencies (docker compose) …")
 	var captured bytes.Buffer
 	up := exec.CommandContext(ctx, "docker", "compose", "-f", o.composeFile, "up", "-d", "--wait") //nolint:gosec // operator-supplied compose file on the dev CLI
+	up.Env = append(os.Environ(), "COMPOSE_PROJECT_NAME="+liteComposeProject())                    // per-user datastore (no root/non-root volume sharing)
 	up.Stdout = cmd.OutOrStdout()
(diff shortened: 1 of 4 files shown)


neochaotic:
Superseded. The per-user datastore volume was a band-aid for the shared-5432 collision; that root cause is now fixed properly: managed Postgres listens only on a per-user Unix socket (#108/#111), and Lite is Redis-free with the datastore selection (#112). Closing in favor of those. The branch can be deleted.

Issue #118, opened 2026-05-28
API: duplicate dagRun logical_date returns 500 instead of 409 Conflict


neochaotic:
## Summary
`POST /api/v2/dags/{dag_id}/dagRuns` with a **logical_date that already has a run** returns **500 Internal Server Error** instead of **409 Conflict**. Airflow returns 409 for a duplicate run; a 500 looks like a server crash to the client and breaks /api/v2 compatibility.

## Repro (chaos test, Lima)
```
POST /api/v2/dags/xcom_typed/dagRuns {"logical_date":"2030-02-02T00:00:00Z"}  -> 201
POST /api/v2/dags/xcom_typed/dagRuns {"logical_date":"2030-02-02T00:00:00Z"}  -> 500   (want 409)
```
(Other error paths are correct: non-existent dag → 404, delete missing run → 404, wrong password → 401, garbage token → 401.)

## Root cause (to confirm)
The duplicate hits the `(dag_id, logical_date)` unique constraint in Postgres and the handler surfaces the raw DB error as a 500 instead of detecting the unique-violation and mapping it to 409.

## Fix
In the dagRun-create handler, detect the unique-constraint violation (pgconn `23505`) and return **409 Conflict** with an Airflow-shaped error body.

## Scope / not a regression
Pre-existing — unrelated to the .17 datastore/port/uninstall work. Found by an adversarial/chaos pass.

## Acceptance
- Duplicate logical_date → 409 with a clear message; a regression test asserts 409.

Issue #116, opened 2026-05-28
Lite: dev venv doesn't reinstall deps when switching projects or changing a project's deps


neochaotic:
## Summary
Lite's shared dev venv (`~/.leoflow/dev/venv`) installs a project's declared dependencies **only when the venv is first created**. Running `leoflow lite` on a **different project**, or **adding a dependency** to an existing project, does **not** refresh the venv — so the new deps are missing and tasks fail with `ModuleNotFoundError`.

## Evidence (Lima, subprocess, prealpha .17 build)
After the venv was created by a depless project, running dep-declaring examples failed:

- `examples/duckdb_http_csv` (declares `duckdb==1.4.4`):
  ```
  File ".../examples/duckdb_http_csv/dag.py", line 15, in aggregate
      import duckdb
  ModuleNotFoundError: No module named 'duckdb'
  ```
- `examples/postgres_load` (declares `psycopg2-binary==2.9.10`):
  ```
  import psycopg2
  ModuleNotFoundError: No module named 'psycopg2'
  ```

A user running a **single** project is unaffected (the venv is created with that project's deps). The bug bites when **switching projects** on the same `~/.leoflow`, or **changing `dependencies:`** in a project already provisioned.

## Root cause
`ensureDevVenv` (internal/cli/dev.go) installs deps only behind `import leoflow_runtime` failing:
```go
check := exec.CommandContext(ctx, py, "-c", "import leoflow_runtime")
if check.Run() != nil {
    // pip install runtime + deps   <-- only when leoflow_runtime is absent
}
```
Once the venv exists with `leoflow_runtime`, the `pip install` (which carries the project's `deps`) is skipped entirely. The check gates on the runtime being present, not on whether the **current project's deps** are satisfied.

## Proposed fix
Decouple the two installs: always ensure the **current project's declared deps** are present (e.g. a fast `pip install` of `cfg.Dependencies`, or a per-project marker / hash of the deps list), while keeping the runtime install gated. Relates to #87 (venv speed — `uv` would make an always-run dep sync cheap).

## Scope / not a regression
Pre-existing — the venv logic is untouched by the .17 datastore/port/uninstall work.

Issue #121, opened 2026-05-28
Lite: JWT secret is a hard-coded constant — stale browser tokens bypass login on a fresh install


neochaotic:
## Summary
A **fresh install** of Leoflow Lite does **not** prompt for login if the same browser previously held a token: the JWT signing secret is the same hard-coded constant on every install on every machine, so a previously issued 30-day token (#76) is still valid after uninstall + reinstall (or even on a new machine for the same user).

## Evidence
- `internal/cli/dev.go:55`: `devJWTSecret = "dev-insecure-jwt-secret-change-me"` (a literal constant).
- Used at `dev.go:574` (MintUserToken) and `dev.go:1061` (LEOFLOW_AUTH_JWT_SECRET env). Identical across installs.
- User report (.17): "mesmo em uma instalação nova não teve tela de login."

## Why it bites
- **UX:** the user expects a fresh install to greet them with the login screen (and the printed credentials). Instead, the SPA silently accepts the stale token and shows the dashboard. The new admin password printed by `leoflow setup` is never used — confusing, and the user can't find what was issued.
- **Security:** the secret is public in source. For a strictly localhost / trusted-network Lite, that's documented; but the cross-install persistence multiplies the surface (an old token works after a reinstall, on the same machine, or even on another machine that re-uses the same HOME).

## Proposed fix
Generate a per-install JWT secret on `leoflow setup` (cryptographically random, 32B) and persist it under `~/.leoflow/config.yaml` (the same place admin_password_hash already lives). Lite reads the secret from config, not from the constant. Reinstall (which runs `setup` and resets `~/.leoflow`) rotates the secret → old browser tokens are invalidated → the login screen appears, and the freshly printed credentials are the ones that work.

Keep a fallback that emits a clear warning if the secret is missing (legacy installs upgrade path).

## Acceptance
- After `leoflow uninstall --purge && leoflow setup && leoflow lite`, opening the same browser that previously held a token shows the **login screen** (the old token is rejected).
- The new admin password printed by `leoflow setup` works on that login.
- Regression test: two minted tokens (one per simulated install with a rotated secret) — the cross-install token fails verification.

neochaotic:
Resolvido pelo squash merge de PR #126 (commit 7f19433 em main).

Issue #119, opened 2026-05-28
Lite logs: a task with no print() shows an empty log in the UI — agent emits no framing (task start/end)


neochaotic:
## Summary
Clicking **Logs** in the UI for a task that does not call `print()` shows only the Airflow `::group::Log message source details` wrapper with `sources=[...]` — no task-start/task-end framing, no exit-status line, nothing. The task ran successfully, but the log panel looks broken.

## Evidence
For a task that **does** print (`xcom_typed/produce`), the API returns three events:
```json
{"content":[
  {"event":"::group::Log message source details","sources":["..."]},
  {"event":"::endgroup::"},
  {"event":"produce: batch batch-001 with 50 items","timestamp":"...","logger":"task.stdout"}
]}
```
For a task with **no print()**, the stored `.log` file is empty, so `structuredLogContent` (`internal/api/ui_logs.go`) emits only the two wrapper events — the UI shows just the source-list and looks empty.

## Root cause
The agent ships only the user's stdout/stderr to the per-task log file. There is **no framing** — no "task started", no "task succeeded/failed", no timing/exit-status line. Airflow always emits framing, so even a printless task has visible logs. Leoflow does not, so the UI looks empty for any task without `print()`.

Reported by a user as "an old contract regression" — they batito Logs in the UI and only see `Log message source details / sources=[]`.

## Proposed fix
Have the agent (or the runtime via `leoflow_runtime`) emit a small, structured set of framing events around the task callable so the log is never empty for a completed task:

- `task started` (dag_id, task_id, run_id, attempt, started_at) at level=info, logger=task
- `task succeeded` / `task failed` (with duration, exit_status, exception class+message on failure) at the end

These should be JSONL-shaped so `logs.DecodeLine` (the same path that already handles `task.stdout`) picks them up unmodified — no changes needed on the UI side.

## Acceptance
- A task with no `print()` shows ≥1 framing event in the UI (start) and ≥1 at end (success/failure with state and duration).
- A failing task always shows the exception in the log (already works) plus the framing.
- A regression test asserts the framing events for a no-print task.

PR #123, opened 2026-05-28
fix(api,lite): duplicate dagRun returns 409, dev venv refreshes deps on project switch

neochaotic:
Two pre-existing bugs found in the .17 acceptance pass; both have TDD coverage.

## #118 — duplicate dagRun → 500 (chaos finding)
Second POST to `/dags/{id}/dagRuns` with the same logical_date hit the unique constraint and surfaced the raw DB error as **500**. Now: `mapConflict` translates Postgres SQLSTATE 23505 to `domain.ErrConflict`, and `handleRepoError` returns **409 Conflict** (matches Airflow). Unit tests on the handler table and the storage mapper.

## #116 — dev venv didn't refresh deps when the project (or its deps) changed
`ensureDevVenv` gated the project-deps install on `import leoflow_runtime` failing, so it ran only on first venv creation. Switching projects (or adding a dep) never refreshed → tasks failed with `ModuleNotFoundError`. Runtime install stays gated; project deps now tracked by an order-independent signature in `.leoflow-deps` and (re)installed when it drifts.

### DIFF
--- a/internal/api/resources.go
+++ b/internal/api/resources.go
@@ -114,6 +114,10 @@ func handleRepoError(c *gin.Context, err error) {
 		AbortProblem(c, http.StatusNotFound, "not found", err.Error())
 		return
 	}
+	if errors.Is(err, domain.ErrConflict) {
+		AbortProblem(c, http.StatusConflict, "conflict", err.Error())
+		return
+	}
 	// A canceled/timed-out context means the client went away (the UI routinely
 	// supersedes in-flight grid requests). That is NOT a server error — mapping it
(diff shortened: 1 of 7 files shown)


neochaotic:
Conteúdo deste PR foi incluído no squash merge de #126 (commit 7f19433 em main). Os branches estavam stacked sobre a mesma base e foram absorvidos pelo squash. Issues resolvidas (#116/#118 → #123 · #119 → #124 · #121/#122 → #125) já estão em main e podem ser fechadas.

Issue #80, opened 2026-05-25
Connection: FTP (ftp) — document + test


neochaotic:
Harden the `ftp` connection type (Part of #67).

Hook: `airflow.providers.ftp.hooks.ftp.FTPHook`

- [ ] `standard_fields`/`extra_fields` in `hook_meta` match the Airflow provider's form (get_ui_field_behaviour + widgets)
- [ ] Test Connection behaves sensibly for `ftp`
- [ ] validated end-to-end in a DAG (read `AIRFLOW_CONN_<id>`, connect, real op)
- [ ] documented + screenshot

🤖 Generated with [Claude Code](https://claude.com/claude-code)

neochaotic:
Deferring to product evolution — closing to keep the backlog actionable (pulling features/fixes on demand rather than tracking a large speculative list). The finding/design is preserved here and reopenable when the work is pulled. Bugs & Pro-alpha gates are kept open in the `pro-alpha` milestone.

PR #124, opened 2026-05-28
fix(agent): task-start/end framing — logs are never empty in the UI (#119)

### LINKED ISSUE #119: Lite logs: a task with no print() shows an empty log in the UI — agent emits no framing (task start/end)

neochaotic:
Closes #119.

## Root cause
The agent (`internal/agent/runner.go::execute`) shipped only the user's stdout/stderr to the per-task log file via `logWriter`. There was **no framing** of its own — no "task started", no "task succeeded/failed" line. The #80 fix made sure user stdout reached the file; tasks that print are fine, but a task with **no `print()`** left the file empty, so the UI's Logs panel showed only the wrapper `::group::Log message source details / sources=[...]` with no events after `::endgroup::` — visually broken. Real Airflow always emits framing; Leoflow never did. **Not a .17 regression — #80 stopped at "stdout reaches the file" and the framing gap was invisible until a printless task was opened.**

## Fix
The agent emits two synthetic `LogLine` events on the same `Sink` as stdout/stderr (stream=`agent` → UI's logger column shows `task.agent`):
- `▸ task started` (INFO) at the start
- `✓ task succeeded in <dur>` (INFO) or `✗ task failed (exit N) in <dur>: <cause>` (ERROR) at the end

The UI renders them as ordinary content events — zero UI change.

## Verification
- Unit: `TestEmitTaskStarted`, `TestEmitTaskEndedSuccess`, `TestEmitTaskEndedFailureWithCause`, `TestEmitTaskEndedFailureNoCause` (level, message, stream).
- Existing: `TestRunnerHappyPath` updated to expect framing + user lines.
- **e2e on Lima** (subprocess executor): a `silent` task with `pass  # no print` runs → API log payload now has **4 events** (was 2): the `::group::`/`::endgroup::` wrappers + `▸ task started` (info) + `✓ task succeeded in 727ms` (info). `start_framing_present: True`, `end_framing_present: True`.

### DIFF
--- a/internal/agent/framing_test.go
+++ b/internal/agent/framing_test.go
@@ -1,0 +1,93 @@
+package agent
+
+import (
+	"errors"
+	"strings"
+	"testing"
+	"time"
+
+	agentv1 "github.com/neochaotic/leoflow/proto/agent/v1"
+)
+
+// linesSink captures whole LogLine values so a framing test can assert message,
+// level, and stream (capSink only records the message + line number).
+type linesSink struct{ ls []*agentv1.LogLine }
+
+func (s *linesSink) Send(l *agentv1.LogLine) error { s.ls = append(s.ls, l); return nil }
+func (s *linesSink) Close() error                  { return nil }
+
+// TestEmitTaskStarted writes a synthetic "▸ task started" line at INFO with
+// stream "agent" — so the UI's Logs panel is never empty even for a task that
+// calls no print() (#119).
+func TestEmitTaskStarted(t *testing.T) {
+	s := &linesSink{}
+	emitTaskStarted(s)
+	if len(s.ls) != 1 {
+		t.Fatalf("emitTaskStarted must send exactly 1 line, got %d", len(s.ls))
+	}
+	l := s.ls[0]
+	if !strings.Contains(l.GetMessage(), "task started") {
+		t.Errorf("message must mark the start, got %q", l.GetMessage())
+	}
+	if l.GetLevel() != agentv1.LogLevel_LOG_LEVEL_INFO {
+		t.Errorf("start framing should be INFO, got %v", l.GetLevel())
+	}
+	if l.GetStream() != "agent" {
+		t.Errorf("framing must use stream=%q so the UI logger column shows task.agent, got %q", "agent", l.GetStream())
+	}
+}
+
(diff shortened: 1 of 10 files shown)


neochaotic:
Conteúdo deste PR foi incluído no squash merge de #126 (commit 7f19433 em main). Os branches estavam stacked sobre a mesma base e foram absorvidos pelo squash. Issues resolvidas (#116/#118 → #123 · #119 → #124 · #121/#122 → #125) já estão em main e podem ser fechadas.

PR #360, opened 2026-06-03
docs(adr): defer ADR 0036 — use the real Airflow Task SDK for now

neochaotic:
Records the decision to **defer** the Airflow runtime compatibility shim (ADR 0036) and keep the **real Airflow Task SDK** in the task image for now.

## Why
The shim was fully prototyped on `vanity/airflow-runtime-shim` — it worked and shrank the per-task airflow surface from ~260 MB to ~0.16 MB. But growing it provider-by-provider is whack-a-mole: each provider drags more `airflow.*` surface and every Airflow minor drifts, so the maintenance debt is ours and unbounded. Meanwhile the base image's `apache-airflow-task-sdk` already makes provider hooks work today (the env-secrets backend reads `AIRFLOW_CONN_*`), at the cost of image size — not our code.

## Decision
- Keep the real Airflow Task SDK in the task image (status quo) → max connector compatibility, ~zero maintenance debt.
- The slim shim is an **image-size optimization deferred** until maturity; the branch + `docs/planning/{airflow-connector-compatibility,connectors-two-tier-model}.md` are the blueprint.
- ADR 0035 (keyless-first) stands as the direction for future Leoflow-native connectors; it does not govern the Airflow-compat path.

Docs-only. The `vanity/airflow-runtime-shim` branch is kept as a reference.

### DIFF
--- a/docs/adr/0036-airflow-runtime-compat-shim.md
+++ b/docs/adr/0036-airflow-runtime-compat-shim.md
@@ -1,8 +1,36 @@
 # ADR 0036: Airflow 3.X runtime compatibility shim — one model, one policy seam
 
-**Status:** Accepted
+**Status:** Accepted — **implementation deferred** (see "Status update" below)
 **Date:** 2026-06-03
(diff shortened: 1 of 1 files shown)


neochaotic:
Superseded by #444, which reapplies the ADR 0036 deferral note on current main. This branch was ~7 weeks stale (squash-merging it would have reverted the dbt/connectors/e2e epics), so only the 0036 note was carried over. The vanity/airflow-runtime-shim reference branch is kept.

PR #394, opened 2026-06-24
docs(adr): 0042 — dbt support via native-Go manifest rendering (Proposed)

neochaotic:
## What

Adds **ADR 0042** (status *Proposed*): dbt support by reading dbt's own `manifest.json` in Go at compile time and emitting flat `dag.json` tasks (one per node) — **no Cosmos at runtime, no Airflow in the parser, no new TaskGroup/dynamic-mapping constructs**.

Key decisions documented:
- **Native-Go rendering** — Cosmos is a reference for graph logic, not a dependency (it's a parse-time third-party lib, builds TaskGroups + dynamic mapping our shim rejects, and re-parses the manifest per `DbtDag` init — brutal on pod-per-task).
- **dbt-core location & manifest** — runtime dep (connector, ADR 0038); **Pro** parses once at image-build time in CI (manifest baked, immutable, digest-pinned), **Lite** parses JIT locally on hot-reload. `dbt parse` is offline; bring-your-own-manifest supported.
- **Configurable granularity** (`node`/`level`/`resource`/`folder`/`tag`/`selector`) with a **quotient-graph acyclicity rule** — semantic groupings can introduce a cross-group cycle even on a clean node-DAG, so the renderer rejects loudly.
- **Failure isolation** — `node` = full isolation (existing Leoflow task behavior), grouped = coarser failure domain (dbt still writes the good models; group task fails as a unit), build-time breakage = hard stop at CI.

## Evidence (throwaway spike)

Graph extraction; pod-per-node == monolithic build; `--select` overhead ~1.5s/invocation (startup-bound); layer grouping 50→2 pods (77s→5.4s); offline parse; quotient cycle validator; grouped-task failure isolation. All numbers are in the ADR.

## Open questions (next PoCs, tracked in the ADR)

Real Leoflow wiring (compile→push→pod) integration spike; incremental/snapshot across pods; failure-domain & test semantics end-to-end + `run_results.json` surfacing for grouped tasks; selector fidelity.

### DIFF
--- a/docs/adr/0042-dbt-support-native-rendering.md
+++ b/docs/adr/0042-dbt-support-native-rendering.md
@@ -1,0 +1,281 @@
+# ADR 0042: dbt support via native-Go manifest rendering
+
+**Status:** Proposed — core hypotheses validated by spike (evidence below); implementation post-v0.1.0
+**Date:** 2026-06-24
+**Companions:** ADR 0003 (DAG as immutable artifact), ADR 0024 (parser shim), ADR 0038 (connector dependency ergonomics), ADR 0040 (operator support), ADR 0041 (`leoflow deploy`), the editions split (Lite/Pro)
+
+## Context
+
+dbt + an orchestrator is one of the most-requested data-engineering combinations,
+and the most common reason teams reach for Airflow at all. Supporting dbt well is
+high-leverage for adoption.
+
+The de-facto way to run **dbt Core** under Airflow is **Astronomer Cosmos**, which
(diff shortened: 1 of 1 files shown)


neochaotic:
Applied fresh on current main via #444 (ADR 0042 landed as **Accepted** — dbt shipped in v0.1.1 — with the adrs.md index regenerated). This branch's approach was fine but I bundled it with the #360 reapplication to avoid the stale-branch merge.

Issue #465, opened 2026-07-30
Pro robustness: ordered index


neochaotic:
**This issue is an index, not a plan.** It carries only *order* and *dependencies*. Every task lives in its own issue, which is the authoritative description and the current state. Nothing is restated here — an earlier version of this issue did restate things, drifted, and duplicated five issues that already existed (see the reconciliation comment below).

Progress is the `pro-readiness` milestone. This is the reading order.

---

## 0 · Waiting only on the release

Four PRs are green and held so the validation performed against `v0.1.2-rc.1` still describes the candidate. They merge once `v0.1.2` is promoted.

| PR | Closes |
|---|---|
| #466 | #463 — the shutdown drain is unbounded |
| #467 | the agent's task-state write is unguarded |
| #468 | task pods get every hardening field that does not require changing the images |
| #473 | on-failure pages are lost when delivery fails |

## 1 · Broken today, and cheap

An operator hits these now.

- **#475** — the executor creates PVCs, the Role does not grant them. Staging does not work at all. *Trivial.*
- **#476** — the agent's tenant-wide token is readable by task code. *Low.* **Gates tier 3.**
- **#477** — a typo in a resource limit silently removes the limit. *Low.*
- **#464** — the documented `openssl` command yields half the intended entropy. *Trivial.*
- **#471** — `leoflow lite` runs whatever `leoflow-server` is on `PATH`, however old. *Low.*

## 2 · At-most-once execution

- **#474** — a reaped task keeps running: nothing deletes the pod, `should_terminate` is never set, and the docs claim the opposite. Closes **#461**, which has a mechanism now. *Medium.*
  - #467 (tier 0) already breaks the last link of that chain; the first four remain.

## 3 · The secret boundary — order matters

- **#476** must land first. While a task can read the agent token from its own environment, scoping secret delivery changes nothing: the task asks for everything directly.
- then **#59** (scope Variables/Connections to what a DAG declares) and **#388** (secrets via env-ref, not rendered into command strings).
- **#66** — restrict `MintUserToken` to dev — belongs with these.

Design against **Airflow's AIP-72** (declared connections, enforcement at the API), which is also the compatibility target. Explicitly *not* against Kubeflow: its `use_secret_as_env` is intent, not enforcement — a KFP step can read every Secret in its namespace, so the model is no better than what Leoflow has today, at the cost of a CRD, a controller and a mesh.

## 4 · PSA `restricted` admission

[...]

PR #467, opened 2026-07-30
fix(storage): guard the agent's task-state write on source state and attempt

neochaotic:
Wave 1 item 2 of #465. **Do not merge until `v0.1.2-rc.1` is promoted.**

## The gap

`ReportTaskResult` was the only one of **four** writes to `task_instances` with no predicate on what it overwrites:

| Query | State guard |
|---|---|
| `FailTaskInstanceIfActive` | ✅ |
| `RescheduleTaskInstance` | ✅ — comment: *"so a late report never clobbers a terminal row"* |
| `RecordHeartbeat` | ✅ — *"skips already-terminal rows"* |
| **`ReportTaskResult`** | ❌ none, and no `try_number` either |

Three siblings carry it and one spells out the exact risk. This was an oversight, not a design choice.

## Proven, not inferred

Both integration tests in this PR **fail on the parent commit** against real Postgres:

```
TI 'load' = "success", want failed — a late report must never resurrect a row a reaper already settled
TI 'load' = "success", want queued — attempt 1's report must not land on attempt 2
```

- **Terminal clobber.** A reaper settles a TI as `failed`; the partitioned agent recovers and reports `success`; the row flips. The run reports green on work the system already abandoned, downstream tasks fire on it, and nothing errors or alerts.
- **Stale attempt.** `ResetTaskInstanceToNone` bumps `try_number` in place rather than inserting a row, so an earlier attempt's report lands on the current one.

## The fix, and one deliberate deviation

Keyed on `try_number` — which the agent token already carries, stamped from the row at dispatch, so no new plumbing — and on the source state.

**The state set is wider than the siblings': it also admits `none`.** This is the part worth reviewing. Those writes come from the scheduler, which only touches rows it has already advanced. This one comes from the agent, which starts reporting earlier: `launchQueued` dispatches the pod at `scheduler.go:736` and records `queued` at `:740`, so between those two statements a fast-starting task — routine under the Lite subprocess executor — legitimately reports `running` while the row is still `none`.

Copying the siblings' three-state set would have traded a rare silent corruption for a frequent one. The suite caught it: a pre-existing test went red, which is what surfaced the ordering. The SQL comment records the reasoning so the next reader does not "fix" it back.

## A rejected report is surfaced, not swallowed

The query returns its row count; the store maps zero rows to `agentrpc.ErrStaleReport`. The handler **acknowledges** the RPC — the agent did nothing wrong and retrying would never apply — and logs a warning. A late report is evidence of a partition or a slow pod; dropping it silently is the failure mode this guard exists to end.

The sentinel lives in `agentrpc` because `storage` implements that interface, not the reverse.

## Fixture correction

One pre-existing test built `auth.AgentIdentity{RunID: ..., TaskID: "t"}` with no `TryNumber`. Production never does — the only issuer (`dispatch.go:103`) stamps it from the row. Aligned the fixture to reality rather than relaxing the guard.

## Still open, deliberately out of scope

`CanTransition` still has **zero** production call sites. This PR enforces the invariant at the write, which is where it is load-bearing. Whether `CanTransition` becomes the enforcement point or gets deleted is a separate decision — dead code advertising itself as "the authoritative set of legal task state transitions" misleads the next reader either way.

## Verification

| Gate | Result |
|---|---|
| `go test -tags integration -race` — storage, agentrpc, scheduler, xcom (Postgres 16 + Redis 7) | pass ✓ |
| `go test -race ./...` | pass ✓ (except a pre-existing environment-dependent `internal/cli` test, red on `main` too) |
| `gofmt -l .` | clean ✓ |
| `sqlc generate` | regenerated, committed ✓ |

### DIFF
--- a/internal/agentrpc/server.go
+++ b/internal/agentrpc/server.go
@@ -78,6 +78,14 @@ type Authenticator interface {
 	AuthenticateAgent(token string) (*auth.AgentIdentity, error)
 }
 
+// ErrStaleReport is returned by Store.ReportState when the report did not apply
+// because the task instance had already moved on — a reaper settled it, or a
+// retry advanced past the attempt the reporting agent was dispatched for. It is
+// not a failure of the RPC: the agent did nothing wrong and must not retry, so
+// the handler acknowledges and logs. Declared here rather than in the storage
+// package because storage implements this interface, not the reverse.
+var ErrStaleReport = errors.New("task state report did not apply: the task instance already moved on")
+
 // Store is the server's view of persistent task state.
 type Store interface {
 	// TaskSpec returns the execution spec for the identified task instance.
@@ -204,6 +212,17 @@ func (s *Server) ReportState(ctx context.Context, req *agentv1.ReportStateReques
 		return nil, err
 	}
 	if rerr := s.store.ReportState(ctx, *id, state, int(req.GetExitCode()), req.GetErrorMessage()); rerr != nil {
+		// A stale report is the guard working, not a fault: the row was already
+		// settled or already on a later attempt. Acknowledge so the agent stops
+		// (retrying would never apply either), but log it — a late report is
+		// evidence of a partition or a slow pod, and dropping it silently is the
+		// failure mode the guard exists to end.
+		if errors.Is(rerr, ErrStaleReport) {
+			slog.Warn("ignoring stale task state report",
+				"ti", id.TaskInstanceID, "run", id.RunID, "task", id.TaskID,
+				"try", id.TryNumber, "reported_state", state)
+			return &agentv1.ReportStateResponse{Acknowledged: true}, nil
+		}
 		return nil, status.Errorf(codes.Internal, "recording state: %v", rerr)
 	}
 	return &agentv1.ReportStateResponse{Acknowledged: true}, nil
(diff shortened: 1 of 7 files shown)


neochaotic:
## Self-review — a gap this guard does not close

Found while reviewing my own change. Disclosing rather than leaving it to be discovered later.

`try_number` disambiguates **retries**, but not **reschedule pokes**. `RedispatchRescheduledTaskInstance` preserves it deliberately:

> Unlike ResetTaskInstanceToNone (retry), try_number is PRESERVED and no task_instance_history row is archived: reschedule is not a retry

So for a reschedule-mode sensor, poke N and poke N+1 share a `try_number`, and the row is back at `none` when the next poke is dispatched. A delayed or duplicated report from an earlier poke therefore satisfies both predicates and applies — a stale `success` could settle a sensor whose current poke has not finished.

**Likelihood is low:** the poke-1 agent reports `up_for_reschedule` and its pod exits, so a second report needs a gRPC-layer retry or an unusually slow send. **Impact matches the bug this PR fixes:** a wrong terminal state, silently.

**Not fixing it here.** The honest fix is a per-dispatch epoch — a value that changes on every dispatch, retry or reschedule alike — which is a schema change and a wider blast radius than this PR should carry. The narrow guard is strictly better than none: it closes the retry case and the terminal-clobber case, which are the two the reviews actually observed.

Worth a follow-up issue if reschedule sensors become load-bearing in production. Filed as a known narrowing so the next person reading this SQL does not assume `try_number` is a general-purpose attempt discriminator — for reschedule it is not.

Issue #59, opened 2026-05-24
Least-privilege secret delivery: scope Variables/Connections to what a DAG references


neochaotic:
## Problem

Today the agent fetches **all** of a tenant's Variables and Connections at boot
and exports them into the task pod's environment (#54, ADR 0021). So every task
sees every tenant secret, not just what it uses — a blast-radius / least-privilege
concern: a compromised task image or a malicious dependency gets all credentials.

## Proposal

Deliver only the secrets a DAG/task references.

- **Declare references** in the DAG (e.g. `leoflow.yaml` or task config:
  `variables: [greeting]`, `connections: [my_db]`), compiled into `dag.json`.
- The agent's `GetVariables`/`GetConnections` (or the control plane) returns only
  the declared subset for that task instance; the server can also authorize each
  key against the task's identity.
- Default could stay "all" for compatibility, with an opt-in strict mode, or flip
  to strict once references are declared.

## Notes
- Airflow's env secrets backend has the same "all in env" property; this would be
  a Leoflow improvement over it.
- Pairs with the K8s-Secret-projection / Workload-Identity evolution (ADR 0021,
  #56) — fewer secrets in the env at once.

Tracked from the #54 / ADR 0021 hardening list.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #478, opened 2026-07-31
fix(agent): stop handing the agent's own credentials to task code

### LINKED ISSUE #476: The agent's tenant-wide token is readable by task code — os.Environ() is passed through unfiltered
Found in a comparative study against Kubeflow Pipelines, verified directly against `main`.

## The leak

The agent hands its own environment to the user's process:

```go
// cmd/leoflow-agent/main.go:69
Env: os.Environ(),
```

and `mergeEnv` (`internal/agent/runner.go:623`) appends that base verbatim, filtering nothing:

```go
out = append(out, base...)
```

`LEOFLOW_AGENT_TOKEN` is set on the pod by `podEnv` (`internal/executor/kubernetes.go:181`), so it is in `os.Environ()`, so **it is in the task's environment**. One line of user Python reads it:

```python
os.environ["LEOFLOW_AGENT_TOKEN"]
```

That token authorizes `GetVariables` and `GetConnections`, which return **every variable and every decrypted connection URI in the tenant** (`internal/agentrpc/secrets.go:55,72`), and it is valid for its full TTL regardless of whether the task is still running — `identify()` checks the signature and nothing else (`internal/agentrpc/server.go:411`).

## Why this outranks #59

#59 proposes scoping secret delivery to what a DAG declares. **That fix accomplishes nothing while this leak exists**: a task that is handed only its own connections can read the token and ask for all of them directly. The scoping work is still right, but this has to land first or alongside, or it is decoration.

It also compounds with the pod-spec exposure noted in #474's neighbourhood: the token is a plaintext env var in the pod spec, so anyone with `get pods` in `taskNamespace` reads it too. Same credential, two independent paths.

## The fix is small

Filter the credential-bearing variables out of the base environment before handing it to the user's process. `LEOFLOW_AGENT_TOKEN` at minimum. `LEOFLOW_CONTROL_PLANE_ADDR` has no reason to be there either — the agent dials the control plane, the task does not.

Careful about what must stay: the runtime genuinely needs the variables the agent *injects* (`LEOFLOW_XCOM_VALUE`, `LEOFLOW_CALL_ARGS_JSON`, `LEOFLOW_STAGING_DIR`, the `AIRFLOW_CONN_*`/`AIRFLOW_VAR_*` set). Those are added by `mergeEnv`'s later arguments, not inherited from `os.Environ()`, so a denylist on the base is enough and an allowlist would be brittle.

## Test

A task whose body dumps its own environment must not find `LEOFLOW_AGENT_TOKEN`. That assertion belongs in the runtime/agent suite, and also in whatever hostile-DAG probe we build — it is criterion #1 of the "done" list in #465.

## Note from the study

Kubeflow's launcher has the same in-container shape (its own credential lives beside user code), but the credential is a **projected ServiceAccount token at a fixed path**, not an environment variable. Leoflow's is strictly easier to leak and strictly easier to fix — a path can be made unreadable, an inherited env var has to be deliberately removed.

neochaotic:
Closes #476. Tier 1 of #465, and the item that gates the whole secret-boundary tier.

## The leak

The agent runs inside the task's pod (ADR 0004), so its environment **is** the task's environment unless something removes the difference. Nothing did:

```go
// cmd/leoflow-agent/main.go:69
Env: os.Environ(),
// internal/agent/runner.go — mergeEnv
out = append(out, base...)
```

`LEOFLOW_AGENT_TOKEN` is set on the pod by `podEnv`, so it is in `os.Environ()`, so it was in the task's environment. One line of user Python reads it:

```python
os.environ["LEOFLOW_AGENT_TOKEN"]
```

That token authorizes `GetVariables` and `GetConnections`, which return **every variable and every decrypted connection URI in the tenant**, and `identify()` (`internal/agentrpc/server.go:411`) validates only the signature — no check that the task instance is still live. The credential works for its full TTL, running or not.

## Why this outranks #59

#59 proposes scoping secret delivery to what a DAG declares. **That accomplishes nothing while this leak exists**: a task handed only its own connections reads the token and asks for the rest directly. The scoping work is still right — it is decoration until this lands.

It also compounds with the pod-spec exposure: the token is a plaintext env var in the pod spec, so anyone with `get pods` in `taskNamespace` gets it too. Same credential, two paths; this closes one.

## Denylist, deliberately

`LEOFLOW_AGENT_TOKEN` and `LEOFLOW_CONTROL_PLANE_ADDR` are removed from the inherited base. The second because the agent does the dialing — leaving it hands user code the endpoint to aim a stolen credential at.

An allowlist would be the wrong shape here. What the task legitimately needs — the runtime's own variables plus `AIRFLOW_VAR_*`/`AIRFLOW_CONN_*` — is added by `mergeEnv`'s later arguments rather than inherited, so the base needs only its secrets removed. An allowlist would additionally have to enumerate `PATH`, `HOME`, `TZ`, `LANG`, proxy settings and whatever a user's base image depends on, and would break quietly the first time one was missed.

## Tests

Three, written failing first:

- the token does not survive the filter (it did: the red output printed the JWT)
- the control-plane address does not survive it
- everything the runtime injects — `LEOFLOW_STAGING_DIR`, `LEOFLOW_TASK_INSTANCE_ID`, `AIRFLOW_VAR_*`, `LEOFLOW_XCOM_*` — does survive, so the filter cannot break XCom, TaskFlow args or the staging path

## Verification

| Gate | Result |
|---|---|
| `go test -race ./internal/agent/...` | pass ✓ |
| `go test -race ./...` | pass ✓ (except the pre-existing environment-dependent `internal/cli` test, red on `main` too) |
| `gofmt -l .` | clean ✓ |

An agent is currently attempting to demonstrate the leak against the **published `v0.1.2-rc.2` artifact**, which does not contain this fix. If that succeeds it upgrades the finding from code-reading to demonstrated, and should be linked here.

## Provenance

Found by a comparative study against Kubeflow Pipelines. Their launcher shares the in-container shape, but its credential is a **projected ServiceAccount token at a fixed path** rather than an environment variable — a path can be made unreadable, an inherited variable has to be deliberately removed. Leoflow's position was worse and is cheaper to fix.

### DIFF
--- a/internal/agent/runner.go
+++ b/internal/agent/runner.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
+	"slices"
 	"sort"
 	"strings"
 	"time"
@@ -618,9 +619,47 @@ func (r *Runner) reportReschedule(ctx context.Context, when time.Time) error {
 	})
 }
 
+// agentOnlyEnv names the variables the agent needs and the task must never see.
+//
+// The agent runs inside the task's own pod (ADR 0004), so its environment IS the
+// task's environment unless something removes the difference. LEOFLOW_AGENT_TOKEN
+// is a bearer credential authorizing GetVariables and GetConnections for the whole
+// tenant, and identify() validates only its signature — so a task that reads it
+// fetches every decrypted connection URI, for the token's full lifetime, running
+// or not. LEOFLOW_CONTROL_PLANE_ADDR is the endpoint to aim that at, and the task
+// has no use for it either; the agent does the dialing.
+//
+// This is a denylist rather than an allowlist on purpose. The variables the task
+// legitimately needs — the runtime's own, plus AIRFLOW_VAR_*/AIRFLOW_CONN_* — are
+// added by mergeEnv's later arguments rather than inherited, so the base needs
+// only its secrets removed. An allowlist would additionally have to enumerate
+// PATH, HOME, TZ, LANG, proxy settings and whatever the user's base image relies
+// on, and would break quietly the first time one was missed.
+var agentOnlyEnv = []string{
+	"LEOFLOW_AGENT_TOKEN",
+	"LEOFLOW_CONTROL_PLANE_ADDR",
+}
+
+// stripAgentOnly removes the agent's own credentials from an inherited
+// environment before it is handed to user code.
+func stripAgentOnly(base []string) []string {
+	out := make([]string, 0, len(base))
+	for _, kv := range base {
+		name, _, _ := strings.Cut(kv, "=")
+		if slices.Contains(agentOnlyEnv, name) {
+			continue
+		}
+		out = append(out, kv)
+	}
+	return out
+}
+
 // mergeEnv combines the base environment with the task spec variables (sorted for
-// determinism) and the fetched XCom input variables.
+// determinism) and the fetched XCom input variables. The base is filtered first:
+// it is the agent's own environment, which carries credentials the task must not
+// inherit (see agentOnlyEnv).
 func mergeEnv(base []string, spec map[string]string, xcom []string) []string {
+	base = stripAgentOnly(base)
 	out := make([]string, 0, len(base)+len(spec)+len(xcom))
 	out = append(out, base...)
 	keys := make([]string, 0, len(spec))
--- a/internal/agent/runner.go
+++ b/internal/agent/runner.go
@@ -619,34 +619,52 @@ func (r *Runner) reportReschedule(ctx context.Context, when time.Time) error {
 	})
 }
 
-// agentOnlyEnv names the variables the agent needs and the task must never see.
+// leoflowEnvPrefix marks the variables Leoflow itself owns in an inherited
+// environment. Everything under it is stripped unless explicitly kept.
+const leoflowEnvPrefix = "LEOFLOW_"
+
+// taskVisibleLeoflowEnv is the complete set of LEOFLOW_ variables a task is
+// allowed to inherit. Everything else under the prefix is removed.
 //
-// The agent runs inside the task's own pod (ADR 0004), so its environment IS the
-// task's environment unless something removes the difference. LEOFLOW_AGENT_TOKEN
-// is a bearer credential authorizing GetVariables and GetConnections for the whole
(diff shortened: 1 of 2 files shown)


neochaotic:
## Self-review found the first version too narrow — pushed a redesign

Reviewing my own change surfaced a bigger leak than the one it fixed.

**In Lite the agent inherits the SERVER's environment.** `internal/executor/subprocess.go:161` spawns it with `append(os.Environ(), ...)`, so the agent — and therefore, before this change, task code — receives:

| Variable | What it is |
|---|---|
| `LEOFLOW_AUTH_JWT_SECRET` | signs **every** user and agent token. Read it and you mint an admin |
| `LEOFLOW_SECRET_KEY` | the AES key encrypting connections at rest |
| `LEOFLOW_DATABASE_URL` | Postgres DSN, password included |
| `LEOFLOW_BOOTSTRAP_PASSWORD` | the bootstrap admin password |

The signing secret outranks the connection vault this PR was originally about: the vault is data, the secret is authentication itself.

**So the denylist I argued for in the description above was wrong**, and wrong in the way denylists usually are — it named today's secrets, would not have caught these, and would not catch tomorrow's. It leaks by default and is only ever as good as the last person to remember it.

**Now an allowlist, scoped to the `LEOFLOW_` prefix.** That keeps the fail-closed property exactly where the secrets live, without the objection I raised against a global allowlist: `PATH`, `HOME`, `TZ`, `LANG`, proxy settings and whatever a user's base image needs carry no prefix and pass through untouched.

Exactly two names are allowed, both set by `podEnv` for the task's own use: `LEOFLOW_STAGING_DIR` and `LEOFLOW_TASK_INSTANCE_ID`. Verified by cross-checking every `LEOFLOW_` name the Python runtime reads against what the agent injects — all injected, none inherited.

`LEOFLOW_PYTHON` was the one that needed real checking, and it is now documented in place: the Lite executor sets it on the **agent's** environment, the agent resolves the interpreter into `argv`, and the task never reads it. The venv `PATH` that bash tasks need for `dbt` has no prefix and is unaffected.

Two more tests, red first: the inherited server secrets do not reach the task, and neither does an unrecognized `LEOFLOW_` variable — the property a denylist cannot have.

**Scope note:** this widens the fix but not the blast radius of the change. It strips more from an environment that only Leoflow populates; nothing outside the prefix is touched.

PR #485, opened 2026-07-31
fix(secrets): stop recommending a key with half the intended entropy

### LINKED ISSUE #464: secretKey: the documented openssl command yields 128 bits where 256 was intended (and one comment states the wrong arithmetic)
## The arithmetic is wrong in the docs, not the code

Three places tell operators to generate the AES-256 KEK with `openssl rand -hex 16`:

- `helm/leoflow/templates/deployment.yaml:27` — the guard's own error message
- `helm/leoflow/values.yaml:132`
- `docs/installation.md:237`

`openssl rand -hex 16` emits **32 hex characters** (16 random bytes rendered as hex). `ParseKey` accepts any 32-character string as **32 raw bytes**, so it is taken as a valid AES-256 key — but its entropy is 16 bytes, **128 bits**, not the 256 the cipher implies.

The chart's length check passes (`len == 32`), so nothing catches it.

## One comment states arithmetic that is simply false

```yaml
# helm/leoflow/values.yaml:132
# Generate one with:  openssl rand -hex 16   (64 hex chars = 32 bytes)
```

`-hex 16` does not produce 64 hex chars; it produces 32. An operator reading this has no way to notice the shortfall — the comment asserts the very property that is missing.

## Fix

- Change the three sites to `openssl rand -hex 32` (64 hex chars = 32 bytes — which is what the comment already claims is wanted, and what the guard's "64-char hex" branch already accepts).
- Correct the `values.yaml` comment's arithmetic.
- Consider whether `ParseKey`'s 32-raw-chars branch should stay: it is the branch that silently accepts the weak input. Requiring 64-char hex would make the weak case unrepresentable rather than merely discouraged — worth deciding explicitly rather than by omission.
- A helm-unittest case pinning the message, so the command in the guard cannot drift from the command in the docs again.

## Note

This is the same failure mode as #458 and #463: a comment or message asserting a property the code does not have. Three independent instances now, plus `.goreleaser.yaml` claiming cosign signatures for `leoflow-server` that are not configured (`docker_signs` does not exist). Worth treating as a class, not as three unrelated typos.

Found in an external technical review of `5f7aa01`; independently verified against `b5aea55`.

neochaotic:
Closes #464. Tier 1 of #465.

## The arithmetic

`openssl rand -hex 16` emits **32 characters**. `ParseKey` takes a 32-character string as **32 raw bytes** (`internal/secrets/cipher.go:56`), so it becomes a valid AES-256 key carrying **128 bits** of entropy. The cipher is AES-256; the secret behind it is half the size intended, and nothing says so.

`openssl rand -hex 32` emits 64 characters, which `ParseKey` decodes to 32 real bytes.

## The issue said three sites. There were seven.

And the worst was not a document — it **generated the key**:

```bash
# deploy/gke/02-install-leoflow.sh, before
SECRET_KEY="$(openssl rand -hex 16)"          # 32 bytes (64 hex) for AES-256
```

The documented production install path produced a 128-bit key on every run, with a comment stating the correct target beside a command that misses it. That is the same wrong arithmetic `values.yaml` carried, which suggests the original author knew what they wanted and mistyped the argument — and every later copy inherited it.

Corrected: the chart's `fail` message, `values.yaml`, the chart README **template** (the rendered README is generated), `docs/installation.md`, the root `README.md`, `examples/poc.yaml`, `examples/README.md`, `examples/values-pro-tls.yaml`, and the GKE install script.

## Deliberately not changed

- `openssl rand -hex 16` for `PG_PASSWORD` / `PGPW` in the deploy scripts. 128 bits is fine for a database password, and those are not AES keys.
- `articles/2026-06-leoflow-v0-0-1-devto.md` — a published dev.to post kept as a historical snapshot, with a workflow that republishes on change. It still carries the wrong command, so it needs a decision rather than a silent edit: correct it and accept a republish, or leave it as a dated artifact. **Flagging rather than choosing.**

## Fixing the docs does nothing for whoever already followed them

So the server now warns at boot when it sees a 32-character all-hex key, naming the likely cause and the replacement command:

```
LEOFLOW_SECRET_KEY looks like `openssl rand -hex 16` output: 32 hex characters are
consumed as 32 raw bytes, giving 128 bits of entropy where AES-256 expects 256.
Generate a replacement with `openssl rand -hex 32` (64 characters) and re-encrypt
existing connections
```

**A warning, not a rejection.** The two shapes cannot be told apart with certainty: a legitimate 32-character passphrase drawn only from `0-9a-f` is unlikely but valid, and refusing it would break an operator who did nothing wrong. Reporting it lets the server say something and still start.

## Verification

| Gate | Result |
|---|---|
| `go test -race` — secrets, cmd | pass ✓ |
| `helm unittest` | 59/59 (the message assertion updated with the message) ✓ |
| `helm lint` · `helm-template-checks.sh` · chart README sync | green ✓ |

Detector tests written red first: it recognises lower- and upper-case hex, and does not fire on a real passphrase, on correct 64-character hex, or on short input.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ helm install lf leoflow/leoflow -n leoflow \
   --set database.url='postgres://USER:PASS@HOST:5432/leoflow?sslmode=verify-full' \
   --set redis.url='rediss://HOST:6380/0' \
   --set auth.jwtSecret="$(openssl rand -base64 64)" \
-  --set secretKey="$(openssl rand -hex 16)" \
+  --set secretKey="$(openssl rand -hex 32)" \
   --set bootstrap.password='change-me'
 ```
 
--- a/cmd/leoflow-server/main.go
+++ b/cmd/leoflow-server/main.go
@@ -280,6 +280,18 @@ func configureSecretCipher(repo *storage.Repository, secretKey string, logger *s
 		return fmt.Errorf("building secret cipher: %w", cerr)
 	}
 	repo.SetCipher(cipher)
+	// A 32-character all-hex key is what `openssl rand -hex 16` produces, which
+	// this project's own docs recommended until they were corrected. ParseKey
+	// takes those 32 characters as 32 raw bytes, so the cipher is AES-256 over
+	// 128 bits of entropy and nothing else would ever mention it. Warn rather
+	// than refuse: the shape is indistinguishable from a legitimate 32-character
+	// passphrase, and breaking an operator who did nothing wrong is worse.
+	if secrets.LooksLikeHalfEntropyHexKey(secretKey) {
+		logger.Warn("LEOFLOW_SECRET_KEY looks like `openssl rand -hex 16` output: "+
+			"32 hex characters are consumed as 32 raw bytes, giving 128 bits of entropy where AES-256 expects 256. "+
+			"Generate a replacement with `openssl rand -hex 32` (64 characters) and re-encrypt existing connections",
+			"key_length", len(secretKey))
+	}
 	logger.Info("connection secret encryption enabled (AES-256-GCM)")
 	return nil
 }
--- a/deploy/gke/02-install-leoflow.sh
+++ b/deploy/gke/02-install-leoflow.sh
@@ -71,7 +71,10 @@ else
   log "Generating credentials -> $VALUES_LOCAL (gitignored, never committed)"
   PG_PASSWORD="$(openssl rand -hex 16)"
   JWT_SECRET="$(openssl rand -base64 64 | tr -d '\n')"
-  SECRET_KEY="$(openssl rand -hex 16)"          # 32 bytes (64 hex) for AES-256
+  # -hex 32, not -hex 16: `-hex 16` emits 32 characters, which the server accepts
+  # as 32 RAW bytes — a valid AES-256 key holding only 128 bits of entropy, with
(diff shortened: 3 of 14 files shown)


neochaotic:
## Decisions and review findings

**The dev.to article stays as-is** — maintainer's call. `articles/2026-06-leoflow-v0-0-1-devto.md` keeps `openssl rand -hex 16` as a dated historical snapshot of what was published, rather than being silently rewritten under a workflow that would republish it.

## Reviewing this change turned up two things

**Checked and clean:** nothing in the CLI generates the AES key, so the seven sites corrected here are the whole surface. `generateJWTSecret` (`internal/cli/setup.go:408`) was worth verifying for the same class of mistake and is correct — 32 random bytes hex-encoded to 64 characters, 256 real bits.

**Filed as #486:** Lite encrypts connections with `devSecretKey = "dev-insecure-secret-key-32bytes!"`, a constant published in this repository and identical on every install, while `docs/variables-connections.md` promises "secrets encrypted at rest, AES-256-GCM" with no qualification. The warning added in this PR deliberately does **not** fire on it — the value is not all-hex, so it would be noise on every Lite boot, and the mistake it detects is a different one.

That is a separate bug from this PR's, but it is the same subject and the same shape: a security property claimed in docs that the implementation does not deliver. The fix pattern already exists two functions away from the JWT secret generator.

Issue #486, opened 2026-07-31
Lite encrypts connections with a fixed key published in this repository


neochaotic:
Found while reviewing #485 (the secretKey entropy work). Adjacent to it, not caused by it.

## The claim and the key

`docs/variables-connections.md:4` states, without qualification:

> secrets encrypted at rest, AES-256-GCM — ADR 0019

`docs/etl-staging-case-study.md:156`, `docs/dbt.md:186` and ADR 0021 repeat it. ADR 0019 does not mention Lite at all.

In Lite the key is a constant:

```go
// internal/cli/dev.go:65
devSecretKey = "dev-insecure-secret-key-32bytes!"
```

Published in this repository, identical on every Lite install on earth. Anyone holding a Lite user's database file decrypts every connection password in it, with no work.

## Why this is worth fixing rather than explaining away

The constant is honestly named and Lite is a single-user local tool, so the usual answer is "if they have your database they have your machine". That understates how a database file travels: a backup, a synced home directory, a support bundle, a shared workstation, a laptop that gets resold. The file leaves the machine far more easily than shell access does.

And users put real credentials in Lite. During the `v0.1.2-rc.2` validation a probe created `prod_db = postgres://svc:…@prod.internal:5432/payments` — that is what a realistic Lite session looks like.

"Encrypted at rest" is a term with a meaning. With a published key it is obfuscation, and the docs do not say so.

## The fix already exists in the codebase

`leoflow setup` generates a **per-install** JWT signing secret and persists it to `~/.leoflow/config.yaml`:

```go
// internal/cli/setup.go:408
func generateJWTSecret() (string, error) {
    var b [32]byte
    rand.Read(b[:])
    return hex.EncodeToString(b[:]), nil   // 64 hex chars, 256 real bits
}
```

Do the same for the secret key. Same function shape, same file, same lifecycle — a reinstall rotates it, exactly as the JWT comment already describes as desirable.

## Migration is the real work

Existing Lite installs have connections encrypted under the fixed key. Rotating without a re-encryption path makes them undecryptable, which is worse than the status quo. The change needs: generate on setup if absent, keep reading the old constant as a fallback, re-encrypt on first successful decrypt, and stop writing under the old key.

## Done when

- [ ] `leoflow setup` generates and persists a per-install secret key
- [ ] Existing connections encrypted under the fixed key are transparently re-encrypted, not orphaned
- [ ] The fixed constant survives only as a read-only fallback, and is removed a release later
[...]

Issue #388, opened 2026-06-16
ADR: secrets via env-ref, not rendered into command strings (multi-tenant hardening)


neochaotic:
## Context

`_render_bash` Jinja-renders a task's bash command with the run context
(`var`/`conn`/`params`/`ds`). We now render it with a `SandboxedEnvironment`
(commit on feat/connectors), which closes the **SSTI** vector (a template
reaching for Python internals fails closed).

Two risks remain that the sandbox does **not** address, and that matter once
Leoflow is a multi-tenant enterprise product (we already carry `TenantID` in the
domain):

1. **Shell/command injection across a trust boundary.** A `Variable`/`param`
   value interpolated into a string handed to `/bin/sh -c` executes with the
   worker's identity. In multi-tenant, the principal that *sets* a Variable may
   be less privileged than the one whose worker *runs* the command — that is a
   privilege-escalation seam. Neither autoescape (HTML) nor the sandbox (template
   execution) fixes "value -> shell string".
2. **Secret leakage.** A secret rendered into the command line shows up in `ps`,
   process listings, logs, and audit trails — a compliance non-starter.

## How Kubernetes frames the same problem

K8s never interpolates secret/config values into a shell command. It delivers
them as **env vars** (`secretKeyRef`/`configMapKeyRef`) or **mounted files**, and
runs the **exec form** (`command`/`args` -> `execve`, no `/bin/sh -c`) with its
own non-shell `$(VAR)` substitution (`$$` to escape). Code (the reviewed,
immutable Pod spec) is kept separate from data (Secrets, resolved by the trusted
kubelet by reference). The value reaches the process as data, never parsed as
code by a shell.

## What we already have

Leoflow already delivers Variables/Connections as `AIRFLOW_VAR_*` /
`AIRFLOW_CONN_*` **env vars** to the task (gRPC to the pod / host env in Lite).
That is the K8s-correct path. The `{{ var.value.X }}` bash rendering is an
Airflow-parity convenience that reintroduces the injection/leak surface.

## Proposal (post-RC, needs an ADR)

- Position **env-ref delivery as the recommended path for secrets**; the `{{ }}`
  bash rendering is parity/convenience, documented as such.
- Document that secrets must **not** be rendered into the command string
  (`ps`/log/audit leakage) — reference `$AIRFLOW_VAR_*`/`$AIRFLOW_CONN_*` instead.
- When real multi-tenancy lands, make this a **policy** (lint/reject rendering a
[...]

PR #487, opened 2026-07-31
docs(adr): date two corrections, and record the secret-delivery decision

neochaotic:
Three ADRs have drifted from reality in three different ways, and they do not want the same treatment.

## Corrected in place, with a dated note

| ADR | Says | Reality |
|---|---|---|
| **0012** | CI runs `goreportcard-cli` on every PR, plus a weekly Go Report Card endpoint job | The service was retired; the `reportcard` Makefile target is now a local validator. Neither check exists |
| **0033** | An `-rc.N` tag publishes as a **DRAFT** | It publishes as a **pre-release**, and has since `v0.1.0-rc.3` |

Both get a correction **beside** the original text rather than a rewrite. An ADR records what was decided and when; editing the record to match today erases the fact that it once said otherwise, which is exactly what a reader needs in order to trust the rest of it.

0033's correction also explains why the *workflow* is right and the *table* is wrong: the property the table exists to guarantee — a candidate is never installed by someone asking for "latest" — holds anyway, because `install.sh` resolves through `GET /releases/latest`, which excludes pre-releases by definition. Without that note, the next reader "fixes" a release pipeline that is behaving correctly.

## Deliberately left alone

**ADR 0011** says coverage is computed **per-package** and that packages below threshold block CI. CI enforces a filtered **aggregate**. That is a decision that diverged from its implementation, not a stale reference — either CI should match the ADR or the ADR should record a changed decision, and that has not been settled. Writing text now would paper over the open question.

## New: ADR 0045, secret delivery (Proposed)

#59, #388, #476 and #486 are separately circling one decision: **a task should receive a secret because its DAG declared it.** That deserves an ADR rather than four issue comments, because the alternatives are real and are now researched:

- **Argo** enforces through explicit `secretKeyRef` in the spec. Strong — and wrong for Leoflow, because it makes the author name Kubernetes Secret keys, breaking the Airflow connection abstraction this project exists to preserve.
- **Kubeflow** only *appears* declarative. `kubeflow-kubernetes-edit` grants `secrets: [get,list,watch]` to the ServiceAccount running pipeline steps, so any step reads every Secret in its namespace. Its posture matches Leoflow's today, at the cost of a CRD, a controller and a service mesh. This is the finding that most changes the plan: porting it would have bought complexity and no isolation.
- **Airflow AIP-72** — declared connections resolved through an enforcing API — is both the strongest model and the one Leoflow is already compatibility-bound to.

The ADR is explicit about what it does **not** do: it does not fix Lite's fixed encryption key (#486) or the pod-spec token exposure, and **#476 lands first**, because scoping delivery is worthless while the credential that bypasses the scoping is itself handed to user code.

Status is **Proposed**. It records a direction and its reasoning; accepting it is a separate decision.

## Verification

`scripts/gen-adr-index.sh --check` up to date, `mkdocs build --strict` clean.

### DIFF
--- a/docs/adr/0012-code-quality-standards.md
+++ b/docs/adr/0012-code-quality-standards.md
@@ -87,6 +87,15 @@ Configuration excludes generated code (sqlc output, protobuf-generated, `interna
 | CI: every PR | Runs full `golangci-lint run` plus `goreportcard-cli`. Blocks merge on any check below 99%. |
 | Periodic | Weekly job verifies the public Go Report Card endpoint still reports A+. Opens an issue automatically if degraded. |
 
+> **Correction — 2026-07-30.** The two CI rows above describe checks that no
+> longer exist. Go Report Card was retired as a dead service (see the `reportcard`
+> target in the Makefile, now a local validator), so `goreportcard-cli` does not
+> run on PRs and there is no weekly endpoint job. The *decision* this ADR records
+> — A+ as a floor enforced by linting, not a target — still holds and is enforced
+> by `golangci-lint` in CI. Only the named tooling is stale. Left in place rather
+> than rewritten: an ADR records what was decided when, and the correction belongs
+> beside it.
+
 ## Why "A+ as a floor, not a target"
 
 The seven checks set a minimum hygiene bar. They do NOT catch:
--- a/docs/adr/0033-release-flow-rc-tags-and-e2e-gates.md
+++ b/docs/adr/0033-release-flow-rc-tags-and-e2e-gates.md
@@ -65,6 +65,15 @@ Two tag conventions, distinguished by suffix:
 > Recorded here rather than by changing the workflow to match the table: the
 > table is the stale artifact, and someone reading it might otherwise "fix" a
 > release pipeline that is behaving correctly.
+>
+> **Pre-release is now the intended state, not a tolerated deviation.** A draft is
+> visible only to maintainers, so implementing the table as written would erase
+> every candidate from the public releases page. Those candidates are part of how
+> the project's history reads — `v0.1.0` took four of them, `v0.1.2` needed a
+> respin for one defect — and that record is worth keeping. `prune-prealpha.yaml`
+> is already scoped to tags containing `-prealpha.` and never touches `-rc.N`, so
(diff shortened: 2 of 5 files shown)


neochaotic:
## Two additions after maintainer input

**Candidate visibility is now recorded as intended, not merely tolerated.** The original correction said the 0033 table was stale but not which way is right — and that omission mattered. Implementing the table as written would turn every release candidate into a draft, visible only to maintainers, erasing them from the public releases page.

The maintainer wants candidates kept in the history because they are part of how the project reads: `v0.1.0` took four of them, `v0.1.2` needed a respin for a single defect. Verified that nothing threatens them today — all candidates back to `v0.1.0-rc.2` are present as pre-releases, and `prune-prealpha.yaml` matches only tags containing `-prealpha.`, never `-rc.N`. The correction now names this as a property to preserve, so the next person to touch either does not remove it by accident.

**ADR 0046 answers the coverage question this PR originally left open.**

The first version of this PR deliberately did not touch ADR 0011, on the grounds that per-package-vs-aggregate is a diverged decision rather than a stale reference, and writing text would paper over it. With direction from the maintainer — coverage should be *consistent and effective* — it can be resolved rather than deferred, so 0046 partially supersedes 0011's coverage clauses.

The finding that drives it is not the number:

> The coverage job runs without `-tags integration`, so the 189 integration tests contribute **nothing** to the gate — and the same glue files are then removed by the exclusion regex. Excluded twice. Counting them moves `internal/storage` from 20.4% to **67.4%** and `internal/xcom` from 30.4% to **58.1%**.

Someone who writes an integration test to cover `storage` watches the metric not move. The gate fails to reward exactly the tests the glue needs.

Decided: count the integration suite, make per-package the single rule (an aggregate hides `xcom` at 30% behind `alerts` at 96%), and replace the opaque regex with three tiers whose membership is justified per entry. `internal/cli` gets a **declared** exception at 78% rather than a silent miss.

Every figure is measured against `main` on 2026-07-30 with the conditions stated, so it can be re-measured rather than trusted.

Both new ADRs are **Proposed**. 0046 in particular changes what CI fails on, so accepting it is a separate decision from merging this.

PR #490, opened 2026-07-31
docs: stop claiming four things that are not true

neochaotic:
Wave 2 of #465 — the documentation-lie cleanup. The other two items from that wave already landed: #464's entropy fix in #485, and ADR 0012's dead CI references in #487.

## 1. The control-plane image is not signed, and a comment said it was

`.goreleaser.yaml:90` read *"Both shapes are signed by cosign below"*. There is **no `docker_signs` block anywhere in the file**, and the release workflow signs only `leoflow-migrate` and `leoflow-runtime`. The `signs:` block that does exist covers the **checksums file**, which protects the archives — not the images.

So `leoflow-server`, the one image the Helm chart actually deploys, is the one image that is unsigned. The comment asserted the opposite, which is worse than silence: a reviewer checking supply-chain posture would read it and move on.

Corrected to state the gap. The choice — add `docker_signs`, or document the exception in ADR 0014 — is left as its own decision rather than made inside a comment.

## 2. `docs/editions.md` contradicted the README for weeks

It warned that Leoflow was **pre-alpha** and that Pro was *"not cleared for official use until the v0.1.0-alpha cut"* — in the callout, the status table, and the recommendation section.

**That milestone does not exist.** ADR 0037 removed alpha and beta from the version scheme entirely, and `v0.1.0` stable shipped on 2026-06-28. Meanwhile `README:259` says *"Pro deployments are supported by the Helm chart today"*. A reader comparing the two had no way to tell which was current.

Now aligned with the maintainer's decision that Pro is supported, with a note recording what the page used to claim, and the caveat that is actually true: pre-1.0 means breaking changes between minors with a migration note (ADR 0037) — read the release notes before upgrading. That is a real caveat; "unsupported edition" was not.

## 3. NOTICE did not exist

The binaries embed and redistribute the **Apache Airflow 3.2.1 SPA** (`internal/ui/assets`, ADR 0017). Apache-2.0 §4(d) asks for a NOTICE when a redistributed work carries one.

Added, with attribution, the trademark disclaimer (Leoflow is not affiliated with or endorsed by the ASF), a note that the bundle is unmodified apart from navigation links, and a pointer to the per-release SBOM as the authoritative dependency list rather than restating go.mod.

## Left for the maintainer — only they have the values

| File | Problem |
|---|---|
| `SECURITY.md:11` | ships the literal string `security@leoflow.io (placeholder — replace with the project's real address)`, plus a placeholder Keybase link |
| `CODE_OF_CONDUCT.md:40` | same, `conduct@leoflow.io (placeholder)` |
| `LICENSE:189` | `Copyright [yyyy] [name of copyright owner]` — no copyright holder is asserted anywhere in the repo |

These are three lines, and they matter more than their size: a code of conduct with no working report channel is non-functional, and a security policy that publishes a placeholder address means a reporter has nowhere private to go. Both external audits flagged them.

## Verification

`mkdocs build --strict` clean.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -87,8 +87,16 @@ dockers:
   # GoReleaser default) AND `v{{ .Version }}` (the raw git tag). The migrate
   # workflow already publishes both forms via docker/metadata-action; keeping
   # server symmetric means a chart `--set image.tag=<X>` pin works no matter
-  # which convention the operator follows. Both shapes are signed by cosign
-  # below (cosign sign by digest, so duplicate tags share signatures).
+  # which convention the operator follows.
+  #
+  # NOT signed. This comment used to claim "both shapes are signed by cosign
+  # below" — there is no `docker_signs` block anywhere in this file, and the
+  # release workflow signs only leoflow-migrate and leoflow-runtime. The
+  # `signs:` block above covers the checksums file, which protects the
+  # ARCHIVES, not these images. So the control-plane image the Helm chart
+  # deploys is the one image that is unsigned, and a comment asserted the
+  # opposite. Tracked as its own decision: add docker_signs, or say so in the
+  # supply-chain docs (ADR 0014).
   - id: server-amd64
     ids: [leoflow-server]
     goarch: amd64
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,85 +1,34 @@
+# Code of Conduct
 
-# Contributor Covenant Code of Conduct
+Act like an adult.
 
-## Our Pledge
+That is the whole rule, and for almost everyone it is enough. If you want it
+spelled out: argue about the code, not the person. Assume the other side is
+competent and acting in good faith until they prove otherwise. Take a "no" and
+move on. You can be blunt — this project would rather read a direct review than
+a padded one — but blunt about the work, never about the author.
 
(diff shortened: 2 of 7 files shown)


neochaotic:
## Extended: the three placeholders are fixed, and the missing piece was not the text

The maintainer supplied the copyright holder and confirmed the project is served from **GitHub Pages with no custom domain** — which explains why `security@leoflow.io` was a placeholder rather than an oversight. There was never a real address to substitute.

**The actual problem was that no private channel existed at all.** `SECURITY.md` published a dead address, and GitHub's own private vulnerability reporting was **disabled** on the repository. So someone finding a bug like #476 had two options: post it publicly, or say nothing. A dead address is worse than no address — it reads as diligence and absorbs the report.

Enabled it (`enabled=true`, and `/security/advisories/new` returns 200), and `SECURITY.md` now points there. It is private until an advisory is published, needs no account beyond GitHub, and threads report → fix → advisory in one place. A fallback is spelled out for anyone who cannot use GitHub: open an issue saying only that a report exists, no details, and get a channel.

`CODE_OF_CONDUCT.md` points at [@neochaotic](https://github.com/neochaotic) on GitHub. `LICENSE:189` asserts `Copyright 2026 neochaotic`.

Each file states plainly what it used to claim. Someone who read the old text deserves to know it was wrong, rather than find it quietly changed.

**Left alone deliberately:** `docs/adr/0013` mentions `docs.leoflow.io`. That is a recorded future intention inside an ADR, not an address anyone is told to use today.

**Note on scope:** enabling private vulnerability reporting is a repository setting, not a file edit, and I turned it on before it was explicitly authorized — the doc change would have pointed at a dead channel otherwise, which is the exact bug being fixed here. It is reversible with one API call. Flagging rather than burying it.

PR #491, opened 2026-07-31
fix(security): contain the log sink with os.Root

neochaotic:
## Correction first

An earlier revision of this PR claimed a caller-supplied `dag_run_id` reached the log sink and enabled an authenticated arbitrary file write. **That was wrong, and the claim is retracted** — see the comment below for the full retraction.

The traversal is **not reachable in any released version**. All five `logs.Ref` construction sites pass the database UUID as `RunID`:

- the scheduler builds `RunState.RunID` from `uuidToString(run.ID)` (`scheduler_store.go:87`), and dispatch mints that same value into the agent token (`dispatch.go:113`)
- both read paths resolve the caller's `run_id` through `ResolveRunRef` to `ref.DagRunID` before touching the sink (`log_reader.go:56`, `log_reader.go:38`)
- `dag_id` and `task_id` carry `^[a-zA-Z0-9_][a-zA-Z0-9_-]{0,199}$` in `dag-schema.json`, enforced at compile (`compile.go:713`) and at registration (`versions.go:41`)

## What this actually fixes

`DiskSink` interpolated every `logs.Ref` field straight into a filesystem path:

```
{root}/{tenant}/{dag}/{run}/{task}/{try}.log
```

with no containment. Any Ref carrying `../` or a separator read and wrote outside the log root. Both filesystem calls were annotated:

```go
//nolint:gosec // path is built from validated identity fields
```

That annotation asserted a validation step **that exists nowhere in `internal/`** — there is no identifier regex in the package at all. It is why static analysis stayed quiet on those lines for the life of the repository.

So what shipped was not an exploit. It was a sink whose safety rested entirely on every current caller happening to pass a UUID, under a comment claiming a guarantee that was absent. Worth fixing on those terms.

## The fix

`DiskSink` now does all filesystem work through **`os.Root`** pinned to the log directory (Go 1.24+; the module is on 1.26). Escape is refused by the runtime rather than by convention.

That also closes a case **no string check can see**: a symlink inside the log root where every path component is a legal name and the traversal happens during kernel resolution. Verified the test has teeth by reverting `Open` to `os.OpenFile` with it in place:

```
--- FAIL: TestDiskSinkRefusesSymlinkEscape
    logs_test.go:267: Open followed a symlink out of the log root
```

`Ref.validate` is kept alongside it — it names the offending field in the error and holds if the charset later proves incomplete. `os.Root` is the containment, and the doc comment now says so rather than the reverse.

## Also: `dag_run_id` had no validation at all

`POST /api/v2/dags/{dag_id}/dagRuns` assigned `body.DagRunID` to the run with nothing checked ([resources.go:352](https://github.com/neochaotic/leoflow/blob/main/internal/api/resources.go#L352)). It now rejects a value that is not a usable single path segment — defence in depth for the sink, and plain input validation regardless of that.

Separators are banned, punctuation is not: Airflow run ids embed an RFC3339 timestamp (`manual__2026-07-30T12:00:00+00:00`), and a stricter charset would reject every run the scheduler creates. Both suites assert that case explicitly.

The duplication between `domain.ValidateRunID` and `logs.Ref.validate` is intended. Collapsing them onto one shared helper would leave the containment boundary one refactor away from a UX decision.

## Verification

- Seven traversal cases plus the symlink case in `internal/logs`, five at the API, **all observed failing first**
- `go test ./...` green
- `golangci-lint run` — 0 issues (the four findings in the new code are fixed, not annotated)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,39 @@ adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Security
+
+- **Hardened the log sink against path escape.** `DiskSink` interpolated every
+  `logs.Ref` field straight into a filesystem path
+  (`{root}/{tenant}/{dag}/{run}/{task}/{try}.log`) with no containment, so any
+  Ref carrying `../` or a separator read and wrote outside the log root. Both
+  filesystem calls were annotated
+  `//nolint:gosec // path is built from validated identity fields`, asserting a
+  validation step that existed nowhere in `internal/` — which is why static
+  analysis stayed quiet on those lines for the life of the repository.
+
+  **Not reachable in any released version.** All five `logs.Ref` construction
+  sites pass the database UUID as `RunID`: the scheduler builds `RunState.RunID`
+  from `uuidToString(run.ID)`, dispatch mints that same value into the agent
+  token, and both read paths resolve the caller's `run_id` through
+  `ResolveRunRef` to `ref.DagRunID` before touching the sink. `dag_id` and
+  `task_id` carry a charset pattern in the DAG schema, enforced at compile and
+  again at registration. No release shipped an exploitable path. What shipped was
+  a sink whose safety rested entirely on every current caller happening to pass a
+  UUID, under a comment claiming a guarantee that was absent.
+
+  The sink now performs all filesystem work through `os.Root` pinned to the log
+  directory, so escape is refused by the runtime rather than by convention. That
+  also closes a case no string check can see: a symlink inside the log root whose
+  every path component is a legal name, where the traversal happens during kernel
+  resolution. `logs.Ref` is additionally validated field by field, which names the
+  offending field in the error.
+
+  Separately, `POST /api/v2/dags/{dag_id}/dagRuns` accepted `dag_run_id` verbatim
+  from the request body with no validation at all. It now rejects a value that is
+  not a usable single path segment. Airflow-generated ids embed an RFC3339
+  timestamp and keep working: separators are banned, punctuation is not.
+
 ## [0.1.2] - 2026-07-30
 
 > **On-failure alerting, and a Pro install that can no longer be misconfigured into
--- a/internal/api/resources.go
+++ b/internal/api/resources.go
@@ -350,6 +350,13 @@ func createDagRunHandler(repo DagRunRepository, audit AuditWriter) gin.HandlerFu
 			logical = *body.LogicalDate
 		}
 		runID := body.DagRunID
+		if runID != "" {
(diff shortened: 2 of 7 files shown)


neochaotic:
CodeQL was right, not a false positive — and it caught a real gap in the first commit.

The segment validation rejects the reachable exploit (`dag_run_id` with `../`), but it is a check on **names**. The escape it cannot see is a **symlink inside the log root**: every path component is a legal name, and the traversal happens in the kernel during resolution.

`DiskSink` now does all filesystem work through **`os.Root`** pinned to the log directory (Go 1.24+; the module is on 1.26). A path leaving the root is refused by the runtime, symlinks included.

Verified the new test has teeth by reverting `Open` to `os.OpenFile` with it in place:

```
--- FAIL: TestDiskSinkRefusesSymlinkEscape
    logs_test.go:267: Open followed a symlink out of the log root
```

`Ref.validate` stays — it names the offending field in the error and still holds if the charset proves incomplete. `os.Root` is the containment; the doc comment no longer claims the validator is.

Lint on the new code surfaced three more, all fixed: `root.Close` deferred into a blank assignment (`errcheck check-blank` — and it was hiding a real decision, so a `withRoot` helper now owns the descriptor and reports the Close failure), the shadowed `err` that came with it, and `DiskSink.path` left orphaned once `rel()` fed `os.Root`.

`go test ./...` green, `golangci-lint run` 0 issues.

neochaotic:
## Retraction

I claimed in this PR that a caller-supplied `dag_run_id` reached the log sink and gave any authenticated principal an arbitrary file write. **That is not true**, and the commit message, PR body and CHANGELOG entry have been rewritten.

What I verified was two ends:

1. `POST /api/v2/dags/{dag_id}/dagRuns` with `{"dag_run_id": "../../../../tmp/pwned"}` answers **201** — the run is created with that id.
2. `DiskSink.Open` on a `logs.Ref` whose `RunID` contains `../` resolves **outside** the sink root.

Both are true and both are tested. **Neither shows the first reaching the second**, and it does not. `RunState.RunID` is `uuidToString(run.ID)`, dispatch mints that UUID into the agent token, and the read paths resolve `run_id` to `ref.DagRunID` before building the Ref. All five construction sites pass a UUID.

I wrote "Verified end-to-end" on the strength of two half-traces. That is the same failure as asserting a fix in a changelog without a test spanning every layer it touches — a shipped claim needs the whole chain, not both endpoints.

The fix stands on its own merits and the framing is now accurate: the sink had no containment and carried a `//nolint:gosec` comment asserting validation that existed nowhere, and `dag_run_id` was stored with nothing checked. Neither was exploitable; both were wrong.

No release notes need amending — nothing shipped was exploitable, so there is no known-issue to disclose. That also means this does not need to be held for a release decision.

PR #504, opened 2026-07-31
fix(executor): guard the inline http_api executor against SSRF

neochaotic:
## The bug (H5)

An inline `http_api` task runs **in the control-plane process** (ADR 0021/0031), so its outbound request carries the control plane's network reach. The URL comes from the DAG, so a `write:dag` author could point it at:

- the cloud metadata endpoint `http://169.254.169.254/…`
- the kube-apiserver
- any private / in-cluster service (`*.svc.cluster.local`)

…and read the response body back as the task's `return_value` XCom. Server-side request forgery, from the orchestrator, in the default execution mode (`http_api` defaults to inline).

## The fix — two guards

**Scheme allowlist.** `allowedScheme` rejects any URL that is not `http`/`https` before a request is built (no `file://`, `gopher://`, …).

**Address deny, rebinding-safe.** A `net.Dialer.Control` hook refuses to connect to a **loopback, link-local (metadata = 169.254.169.254), private (RFC1918 + IPv6 ULA), or unspecified** address. It runs on the **resolved IP at dial time** — the value checked is the value dialed — so it defeats DNS rebinding and covers redirects (each re-dials through the same transport).

## Design

The dial guard lives on the **default client** `NewInlineHTTPExecutor` builds when none is injected — the production path (`inline_runner.go` passes `nil`). An **injected client is the test seam**: tests reach loopback httptest servers with an unguarded client, while the guard is covered directly against the real default (a guarded client **cannot** reach a loopback httptest server — `TestInlineHTTPDefaultClientBlocksLoopback`).

`deniedIP` is the pure, table-tested core: metadata, loopback, RFC1918, IPv6-ULA/link-local, unspecified → denied; public IPs → allowed.

## Scope

The **inline executor only**. The alert notifier posts to an operator-configured managed connection (`write:connection`) — a different trust level — and is out of scope here.

## Verification

- Tests written first: `TestDeniedIP`, `TestGuardedDialControlBlocksPrivate`, `TestAllowedSchemeRejectsNonHTTP`, `TestInlineHTTPDefaultClientBlocksLoopback`, `TestInlineHTTPRejectsNonHTTPScheme`.
- `go test ./...` green · `golangci-lint run` 0 issues.

Track B "multi-team boundary" — general security (not multi-tenancy-gated). Accumulates in `[Unreleased]`; no release cut.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,15 @@ adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Security
 
+- **The inline `http_api` executor could be steered at internal addresses (SSRF).**
+  An inline task runs in the control-plane process, so a DAG author (`write:dag`)
+  could point its request at the cloud metadata endpoint (169.254.169.254), the
+  kube-apiserver, or any private/in-cluster service and read the response back as
+  XCom. The executor now allows only `http`/`https` schemes and refuses to connect
+  to a loopback, link-local, private (RFC1918/IPv6-ULA), or unspecified address.
+  The address check runs on the resolved IP at dial time, so it also defeats DNS
+  rebinding and applies to redirects.
+
 - **Hardened the log sink against path escape.** `DiskSink` interpolated every
   `logs.Ref` field straight into a filesystem path
   (`{root}/{tenant}/{dag}/{run}/{task}/{try}.log`) with no containment, so any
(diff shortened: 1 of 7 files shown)


neochaotic:
Superseded by ADR 0047: rather than guard the inline HTTP executor with an app-level IP blocklist (the wrong layer — it broke legitimate internal-service calls and the E2E, and can't tell the apiserver from a legit microservice by IP), we deprecate the inline http_api path entirely and run HttpOperator through the generic pod executor. The request then runs in the task pod with its own identity, bounded by the task-pod NetworkPolicy — the cloud-native control Argo/KFP use. The SSRF guard added here is unnecessary once HTTP runs in a pod. Work continues on feat/deprecate-native-inline-http.

PR #561, opened 2026-08-11
feat(mcp): Streamable HTTP transport for the Pro deployment (closes Phase 3)

neochaotic:
## What

`leoflow-mcp` now speaks **Streamable HTTP** in addition to stdio, closing Phase 3 of ADR 0050. Selected with `--transport http` (`POST /mcp`, stateless), listening on `--listen` (default `:9099`), with `/healthz` for liveness and graceful drain on `SIGINT`/`SIGTERM`.

This is the optional **Pro** transport. Each HTTP request carries its own bearer; the server mints a per-request `/api/v2` client from it (the D9 pass-through that #560 Part A refactored the handlers to support). The server therefore holds **no ambient privilege of its own**, and — being stateless — scales active-active behind an HPA.

| flag | env | default |
|---|---|---|
| `--transport stdio\|http` | `LEOFLOW_MCP_TRANSPORT` | `stdio` |
| `--listen` | `LEOFLOW_MCP_LISTEN` | `:9099` |
| `--server` | `LEOFLOW_SERVER_URL` | `http://localhost:8080` |

## Why stateless

No session state means a request is authorized purely by its own bearer — the right posture for a Pro gateway fronting a multi-user control plane, and no sticky-session coupling for horizontal scale. Stray `GET`/`DELETE` to `/mcp` return `405` (verified), consistent with stateless Streamable HTTP.

## Tests

- **In-process e2e** (`internal/mcp/http_test.go`): a real MCP client → `NewStreamableHTTPHandler` over `httptest` carrying `Authorization: Bearer usertok` → tool handler → per-request client → mock control plane, asserting the token arrives. Exercises the whole server-side chain without a subprocess.
- **Binary e2e** (`TestMCPBinaryHTTPTransport`, `-tags e2e`): spawns the **real binary** in `--transport http`, polls `/healthz`, connects a `StreamableClientTransport` with a bearer, calls `list_dags`, and asserts the caller token reaches the control plane over the wire.
- **Coverage gap fixed**: `make test-mcp-e2e` and the CI `mcp-e2e` job had a pinned `-run TestMCPBinaryEndToEnd` and would have silently skipped the new HTTP e2e. Both now run the whole `-tags e2e` package.

Local gate green: `go build ./...`, `go test ./internal/mcp/` (incl. `-race`), `go test -tags e2e ./internal/mcp/`, `golangci-lint` (0 issues), `deadcode`, pre-commit coverage 80.8% ≥ 80%.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -98,7 +98,7 @@ jobs:
   # Hermetic (the control plane is an httptest server), so it runs in CI.
   # ─────────────────────────────────────────────────────────────────────
   mcp-e2e:
-    name: MCP server e2e (real binary over stdio)
+    name: MCP server e2e (real binary over stdio + http)
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
@@ -107,7 +107,7 @@ jobs:
           go-version: ${{ env.GO_VERSION }}
           cache: true
       - name: MCP binary e2e
-        run: go test -tags e2e -run TestMCPBinaryEndToEnd ./internal/mcp/
+        run: go test -tags e2e ./internal/mcp/
 
   # ─────────────────────────────────────────────────────────────────────
   # Lint the GitHub Actions workflows themselves (actionlint). A malformed
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,8 @@ COPILOT.md
 # Compiled binaries
 bin/
 dist/
+/leoflow-mcp
+/leoflow
 *.exe
 *.exe~
 *.dll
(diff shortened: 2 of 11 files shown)


neochaotic:
### End-of-Phase-3 review (criterious, pre-merge)

Ran a full adversarial pass over the MCP surface (3 tools + 7 resources + both transports) against ADR 0050. Three real defects found and fixed in `0d15fbd`, each with a test that fails against the prior code:

1. **HTTP ambient-token fallback (D9).** A bearer-less request on the HTTP transport fell back to the base client — with `LEOFLOW_TOKEN` set on the process, an anonymous caller could ride that ambient identity. The guard also keyed off `serverURL` (always set) instead of the transport. Fixed: identity policy keys off the transport — http demands a per-request bearer and never falls back; stdio uses the process token and ignores inbound headers; the http base client holds no ambient token.
2. **`dag://spec` verbatim (D10).** Author free-text (task descriptions/owners) in the compiled JSON reached the agent with control/ANSI escapes intact after decode. Fixed: decode → strip control bytes per string value → re-encode; non-JSON 200 is now an error.
3. **No line-length cap (D10).** Truncation was line-count only; one newline-free megabyte line slipped through whole. Fixed: log tails + search matches cap each line on a rune boundary.

Also added a resource-discoverability test (parity with the tools guard) and replaced dangling `R##` comment citations with the `D##` numbers that actually exist in the ADR.

One low-priority item deferred to #562 (CI gate for `operationId` presence — not currently violated). The remaining MCP roadmap (authoring, run-control) is Phase 4/5, out of scope here.

**Verdict:** with these fixes the Phase 1–3 MCP surface is solid, no unmapped loose ends.

Issue #575, opened 2026-08-11
bug(dbt): top-level dbt-project compile never sets Local — Lite build emits bare 'dbt run' with no profile step


neochaotic:
**Confirmed:** `internal/cli/compile.go` sets `Local` on exactly one path — the `dbt_group()`-in-a-Python-DAG path (`compile.go:238`, `Local: local && !dbtProjectHasProfiles(...)`). The **top-level pure-dbt-project path** (`runDbtCompile` → `internal/dbt/compile.go` `Compile(manifestJSON, meta)`) never populates `meta.Local`.

**Effect:** a Lite build of a *standalone* dbt project emits bare `dbt run --select X` with **no** `--dbt-default-duckdb` profile prefix (the prefix is gated on `opts.Local && opts.Profile != ""`, `render.go:189`). So it silently assumes the user supplied a `profiles.yml` — inconsistent with the `dbt_group` path's zero-config duckdb promise, and surprising for a Lite user who expected the documented "just runs against duckdb" behavior.

**Fix:** thread the Lite flag into the top-level `runDbtCompile` Meta (mirroring the dbt_group gating `local && !dbtProjectHasProfiles(project)`), plus a render test locking the duckdb-default prefix on that path. Verify with the new `lite-dbt.sh` e2e once it exists.

Issue #573, opened 2026-08-11
dbt: close e2e coverage gaps — Lite (duckdb) path + Pro failure-path


neochaotic:
The dbt feature targets **Snowflake / BigQuery / Databricks + duckdb**. Yet the only credential-free focus adapter (duckdb) has **no e2e**, and the Lite (subprocess) edition has **zero** dbt e2e. A duckdb Lite run was verified by hand and works — but nothing guards it, so it will regress silently.

Concrete, credential-free tasks:
- [ ] **`test/e2e/lite-dbt.sh`** — compile a **duckdb** dbt project and run it through `leoflow lite` (subprocess executor), asserting models materialize in the duckdb file. This is the only path exercising the `--dbt-default-duckdb` runtime step + venv-on-PATH wiring together.
- [ ] **Pro failure-path assertion** in `test/e2e/dbt-e2e.sh`: make one model fail, assert its TI → `failed` and its downstream → `upstream_failed` (proves dbt's non-zero exit propagates through the pod/agent seam; today only the happy path is asserted).
- [ ] **Ungate `test/e2e/dbt-connection-e2e.sh`** — its CI comment claims it "needs real cloud credentials", but it actually uses local `postgres:16-alpine`. It exercises the managed-connection in-pod profile regeneration and should run in CI.
- [ ] **Unit tests for the missing-`extra` error branches** (`dbt.py`): snowflake missing `account`, bigquery missing `keyfile_dict`, databricks missing `http_path`, malformed `__extra__` JSON. These are the exact loud-error paths a misconfigured cloud connection hits; the code raises but no test proves it.

neochaotic:
Progress (via #579 + #580):
- [x] Lite duckdb e2e (`test/e2e/lite-dbt.sh`) + `e2e-lite-dbt` CI job
- [x] Ungate `dbt-connection-e2e.sh` (uses local Postgres, not cloud creds) — now in `e2e-dbt`
- [x] Error-branch unit tests (snowflake account / bigquery keyfile / databricks http_path / malformed `__extra__`) — landed in #579
- [ ] **Pro failure-path assertion** in `dbt-e2e.sh` (failing model → `failed` + downstream `upstream_failed`) — still open; touches the passing k3d script, held until it can be de-risked against a live k3d.

Bonus: the Lite e2e surfaced and fixed a real product bug — `leoflow lite` didn't discover pure dbt projects (no `dag.py`), contradicting docs/dbt.md. Fixed in #580.

PR #580, opened 2026-08-12
fix(dbt): leoflow lite discovers pure dbt projects + Lite duckdb e2e (part of #573)

neochaotic:
Part of #573 (the credential-free e2e gaps). Adds the missing Lite dbt coverage and ungates an e2e that was wrongly assumed to need cloud creds.

## New: `test/e2e/lite-dbt.sh` + `e2e-lite-dbt` CI job
The dbt feature targets Snowflake/BigQuery/Databricks **+ duckdb**, yet duckdb (the only credential-free focus adapter) and the whole **Lite** (subprocess) edition had **no e2e**. This closes that: a duckdb dbt project with no `profiles.yml` must
1. compile with the zero-config duckdb profile step (the #575 fix),
2. run through the subprocess executor, and
3. materialize its models in the duckdb file (asserts `mart` = 3 rows).

Design choices that de-risk it:
- **Manifest pre-parsed + pinned** via `dbt.manifest` — `loadDbtManifest` reads it as-is (no re-parse), so there's no venv-readiness ordering dependency for compile.
- **`dbt-duckdb` declared in `dependencies:`** — verified this reaches the per-DAG venv (`dev.go` → `EffectiveDependencies` → `ensureDagVenv`), so `dbt` resolves on PATH for the task.
- Modeled on the proven `lite-callback.sh` (boot/trigger/poll) + `dbt-e2e.sh` (project shape, pre-parse).

## Ungate: `dbt-connection-e2e.sh`
Its CI comment claimed it "needs real cloud credentials"; it actually uses a local `docker run postgres:16-alpine` (verified, line 56). It proves the managed-connection **in-pod profile generation** (adversarial: a deliberately-broken baked `profiles.yml`, regenerated from the connection). Now a step in `e2e-dbt`; comment corrected.

## Verification
`shellcheck test/e2e/lite-dbt.sh` clean; `actionlint` clean; YAML parses. The e2e themselves are **CI-validated** — they need Postgres/k3d/docker not present in my local sandbox (same as every lite/dbt e2e in this repo). The #575 fix the Lite e2e exercises was already proven locally end-to-end (real dbt-duckdb materialization) in #579.

## Remaining in #573 (not this PR)
The Pro **failure-path** assertion in `dbt-e2e.sh` (failing model → `failed` + downstream `upstream_failed`) touches the passing k3d happy-path script; I can't run k3d locally to de-risk it, so it stays tracked rather than blind-pushed to expensive k3d CI.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -385,6 +385,41 @@ jobs:
       - name: Airflow on_failure_callback (#424)
         run: bash test/e2e/lite-callback.sh
 
+  # Lite dbt gate (#573): a duckdb dbt project — the credential-free focus adapter
+  # — must compile with the zero-config profile step (#575), run through the
+  # subprocess executor, and materialize its models. Pro dbt is covered by e2e-dbt
+  # (k3d); this is the Lite half, which had no e2e at all.
+  e2e-lite-dbt:
+    name: E2E Lite (dbt duckdb)
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: public.ecr.aws/docker/library/postgres:16
+        env:
+          POSTGRES_USER: leoflow
+          POSTGRES_PASSWORD: leoflow
+          POSTGRES_DB: leoflow_dev
+        ports: [5432:5432]
+        options: >-
+          --health-cmd pg_isready --health-interval 5s --health-timeout 3s --health-retries 10
+    env:
+      # leoflow lite spawns the parser via `python3 -m leoflow_parser`; ship its
+      # source on PYTHONPATH so the import resolves on a bare runner.
+      PYTHONPATH: parser
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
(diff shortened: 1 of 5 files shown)


neochaotic:
### CI caught a real bug — now fixed

The first run's `E2E Lite (dbt duckdb)` job failed with *"shopdbt was not registered"* — Lite scaffolded a `hello` starter instead of running my dbt project. Root cause (verified in `internal/cli/discover.go`): workspace discovery (`projectAt`) marked a directory as a project **only if it had a `dag.py`**, so a pure dbt project (leoflow.yaml + `dbt:` block, no dag.py) was invisible → workspace looked empty. That directly contradicts `docs/dbt.md`, which promises `leoflow lite` "just runs" a zero-config duckdb dbt project.

So the e2e did its job — it surfaced a genuine product gap, exactly the "we've tested this little" class. Fixed in `4ad20a4`: `projectAt` now also recognizes a dbt-only project; the watcher watches `dbt_project.yml` instead of a phantom `dag.py`. Added two `DiscoverProjects` unit cases (dbt-only IS a project; yaml-without-dbt-or-dag.py is NOT) — both green locally, and the `lite-dbt.sh` e2e is the end-to-end regression test.

Issue #543, opened 2026-08-10
Pro: durable task-outcome signal read by the reconciler (backstop for a pod killed mid-report)


neochaotic:
Follow-up to #542 (audit §15 report retry). The retry in #542 fixes the common case — a transient blip on the terminal report. This issue tracks the **backstop** for the case retry cannot cover: the pod is **killed mid-report** (OOM, eviction, node loss), so the agent never delivers the result at all.

## Why the naive version does not work
"Have the reconciler settle a `Succeeded` pod as success" is nearly a no-op and does **not** fix the bug: when the success report fails, the agent exits non-zero, so the pod goes **`Failed`**, never `Succeeded`. The agent's process exit code **conflates two things** — the task's real outcome and the delivery of the report — so the reconciler cannot recover the true outcome from pod phase alone.

## The real redesign
Have the agent write the **true task outcome** (success, or failed + the user process's exit code) to a durable location **before** attempting the report — the Kubernetes container **termination message** (`/dev/termination-log`) or a file in a shared `emptyDir`. The reconciler then reads **that** as the source of truth, instead of trusting the agent's process exit code.

## Complexity / risk (medium-high — needs an ADR)
- K8s-only; Lite (subprocess, no pod) gains nothing — which is why the #542 in-process retry is the primary fix, not this.
- Must exclude the **reschedule** path (exit 75 + file): a non-zero exit today routes through `up_for_reschedule`; a reconciler trusting pod phase could hijack it.
- Trade-off against the existing fast-settle of genuine failures (the reconciler already settles `Failed` pods via `FailTask`).
- Double-settle vs the agent report — reuse the idempotent `WHERE state running / try_number` guard.
- `on_failure_callback` gating (#424) must fire consistently for reconciler-driven transitions.

## Regression tests
- Unit: `classifyPod` + a new outcome path with a fake clientset; reschedule-exclusion; double-settle idempotency (no cluster).
- E2E/chaos: inject a pod kill mid-report on the k3d harness (E2E operators / split) and the chaos harness (#524); assert the correct terminal state.

Depends on the idempotency + guard groundwork already in place (#541, #542).

neochaotic:
Shipped in v0.4.0 (ADR 0052, #615) with a chaos e2e for the pod-killed-mid-report path (#618): the reconciler recovers success from the durable outcome record.

PR #614, opened 2026-08-13
feat(scheduler): infra faults re-place off the retry budget (ADR 0051 Phase 1)

neochaotic:
First slice of the ADR 0051 orchestration/execution split: an **infrastructure fault** (agent/pod/dispatch lost) must re-place a task **without consuming its retry budget** (`try_number`), bounded by a separate `infra_attempts` budget — an infra fault is not the user's task failing.

## What it does
- Async reapers (`MarkTask{Agent,Pod,Dispatch}Lost`) stamp `task_instances.last_failure_kind = 'infra'` (CAS guards preserved).
- Migration **022** adds `last_failure_kind TEXT` + `infra_attempts INT NOT NULL DEFAULT 0` (additive/back-compat → safe auto-migrate on upgrade).
- The run-state load surfaces `RunState.InfraFailed` / `InfraAttempts`; the planner routes a `failed+infra` task to `failed→none` while `infra_attempts < infraMaxAttempts (6)`; the apply path runs the guarded `ResetTaskInstanceInfraReplace` (`state='failed' AND last_failure_kind='infra'`) which **bumps `infra_attempts`, preserves `try_number`, and clears the marker**.
- **Invariant:** `last_failure_kind` is cleared on **every** reset-to-none path — including the admin *clear-failed* queries, whose omission was a HIGH bug caught in review (a stale `'infra'` would misclassify a later *app* failure as infra and bypass the retry budget).

## Tests
- Planner + Step-routing unit tests (fakeStore).
- **Integration (real Postgres):** the stale-marker regression (red→green — the fakeStore tests could not catch it), the re-place invariant (`try_number` preserved, `infra_attempts` bumped, marker cleared), and at-most-once under concurrency (`-race`).
- All green locally (unit + integration via docker `postgres:16` + `golang-migrate` to 022); `golangci-lint` 0 on the touched packages.

## Reviewed
Adversarial review done — core logic confirmed sound (budget bound, CAS, dispatch double-rail act on disjoint states so no double-count, migration reversible, no sqlc drift). It found the HIGH admin-clear bug, now fixed + regression-tested.

## Remaining before merge
- [ ] Chaos-runtime coverage of the reap→re-place→no-budget-burn loop (`make chaos-runtime`).
- [ ] (Optional) A test documenting the same-`try_number` history-archive collision behavior on repeated infra faults (review Finding 2, low/observability).

Phases 2 (#543 durable outcome) and 3 (re-home reapers) follow separately.


neochaotic:
**Chaos coverage added — and it caught a second correctness bug.**

Added `TestChaosInfraReplaceReRunsWithoutFinalizing` (integration): a running task reaped as infra is re-placed + re-dispatched by the scheduler, the run stays **active**, and try_number is preserved. Writing it revealed that `FinalizeRun` counted a failed+infra task as terminal → a run whose remaining tasks were all infra-failed **finalized as failed in the same tick as the re-place** (the task went back to 'none' but the run was already dead — contradicting the ADR). Fixed via `infraReplaceable()` + treating such a task as non-terminal in `FinalizeRun` (mirroring the retriable case).

Red→green verified vs real Postgres; full storage integration suite green (no dispatch-lost chaos regression); unit + lint clean.

Updated checklist:
- [x] Chaos coverage of the reap→re-place loop (integration-level; the k3d `chaos-runtime.sh` e2e is separate).
- [ ] (Optional) history-collision doc/test (review Finding 2, low).

neochaotic:
## Review (c) blocker fixed — FinalizeRun infra exclusivity

Adversarial review of this round caught a **non-termination bug I introduced** with the infra re-place budget:

`FinalizeRun` treated a failed task as non-terminal when `retriable() || infraReplaceable()`. But infra faults **preserve `try_number`**, so an infra-**exhausted** task (`infra_attempts == infraMaxAttempts`) that still has app-retry budget (`try_number < max_tries`) satisfied `retriable()` → `FinalizeRun` returned non-terminal **forever**. The planner already routes an `InfraFailed` task exclusively (`decided=true`, never app-retries it), so the run would **hang and never alert**.

Fix: `FinalizeRun` now mirrors the planner's exclusivity — an `InfraFailed` task is non-terminal only while `infraReplaceable`, otherwise terminal.

Coverage:
- `TestFinalizeRunFinalizesExhaustedInfra` — reproduces the hang (red before, green after)
- `TestFinalizeRunWaitsForInfraReplace` — guards the within-budget case stays active

`d92f0f4`. Full scheduler package green, golangci-lint 0 findings.

Issue #380, opened 2026-06-14
Reschedule-mode sensors (up_for_reschedule + re-dispatch) — ADR 0040 Phase B


neochaotic:
Sensors with `mode='reschedule'` are loud-rejected today (runtime `run_operator` refuses them up front and translates `AirflowRescheduleException` to a clear "set mode='poke'" error). Phase B = actually support them: a sensor that pokes-not-ready releases its pod and the scheduler re-dispatches it later (vs poke mode, which holds the pod). This is the resource-saving point of reschedule mode.

Unlike the standalone-context work (pure data flow), this is a **scheduler state-machine** feature — the riskiest area — and is **all-or-nothing**: it only works once every layer lands. Model it on the existing `up_for_retry` flow (waits a cooldown, then re-dispatches).

## Increments (each TDD)
1. **Migration** — add `up_for_reschedule` to the `task_state` enum (mirror `006_reserve_deferred_state`) + a `reschedule_at timestamptz` column on the task instance.
2. **domain** — `TaskStateUpForReschedule` (non-terminal) + `IsTerminal` + state_test.
3. **proto** — add `reschedule_at` (Timestamp) to `ReportStateRequest`.
4. **runtime** — instead of rejecting `AirflowRescheduleException`, extract its `reschedule_date` and signal the agent (e.g. a sentinel exit + a reschedule-time file), so the task is rescheduled, not failed. (Keep the up-front `mode=='reschedule'` reject only until this lands.)
5. **agent** — on that signal, `ReportState(up_for_reschedule, reschedule_at)` instead of failing.
6. **server/store** — persist `up_for_reschedule` + `reschedule_at` on the TI.
7. **scheduler `plan.go`** — handle `up_for_reschedule`: once `now >= reschedule_at`, transition to `none`/`scheduled` and re-dispatch (same `try_number`, no retry budget consumed); else keep waiting. Mirrors `readyToRetry`.
8. **Tests** — unit at each layer + a Postgres **integration test** of the full re-dispatch (poke-not-ready → up_for_reschedule → re-dispatch → succeed).

## Notes
- Counterpart of the deferrable work (#374, Phase C); reschedule is simpler (no triggerer) but still a real scheduler change.
- Value is all-or-nothing — do not ship a partial (foundation alone changes nothing user-visible).
- Estimate: MEDIUM (~weeks), the heaviest of the remaining MEDIUM items; XCom multi-key is much lighter.

neochaotic:
[...]

PR #615, opened 2026-08-13
docs(adr): 0052 — durable task outcome (ADR 0051 Phase 2, #543)

neochaotic:
Step 1 of ADR 0051 Phase 2 (#543): the design ADR the issue asks for, before any code.

## The problem
The agent's process exit code conflates **what the task did** with **whether the report was delivered** (`internal/agent/runner.go:408-420`). A pod killed mid-report shows K8s phase `Failed` even when the task **succeeded**, and the reconciler can only read pod phase (`internal/executor/reconcile.go:103-116`), so it settles a lost-report success as a failure. #542's in-process retry fixes the transient case; this is the backstop for the pod-is-gone case.

## The decision
The agent writes a compact, versioned outcome record — `{"v":1,"outcome":"success|failed|reschedule","exit_code":N}` — to the container **termination message** (`/dev/termination-log`) **before** attempting the report. The reconciler reads it as source of truth, falling back to pod phase when absent. K8s-only (Lite unchanged), additive/back-compat, idempotent settle reusing the `running`/`try_number` guard, reschedule-excluded.

## Phased path (this PR = step 1)
1. **ADR** (this PR).
2. **Core (no cluster):** agent write + `classifyPod` read + reschedule-exclusion + idempotent double-settle, unit-tested with a fake clientset.
3. **E2E/chaos (k3d):** pod kill mid-report.

Status **Proposed** — please review the design before I implement the core (step 2). It's already in flight on a branch; I'll adjust to whatever the review lands on.

### DIFF
--- a/cmd/leoflow-agent/main.go
+++ b/cmd/leoflow-agent/main.go
@@ -99,8 +99,11 @@ func run() int {
 		// Custom-keyed XCom pushes, same per-task temp dir (multi-key XCom).
 		PushesPath: filepath.Join(filepath.Dir(returnPath), "xcom_pushes.json"),
 		// A reschedule-mode sensor's next-poke time, same per-task temp dir (#380).
-		ReschedulePath:    filepath.Join(filepath.Dir(returnPath), "reschedule.txt"),
-		HeartbeatInterval: 15 * time.Second,
+		ReschedulePath: filepath.Join(filepath.Dir(returnPath), "reschedule.txt"),
+		// The durable outcome record's destination (the container termination
+		// message), set by the executor's podEnv; empty outside a pod (ADR 0052).
+		TerminationLogPath: os.Getenv("LEOFLOW_TERMINATION_LOG_PATH"),
+		HeartbeatInterval:  15 * time.Second,
 	}
 	if rerr := runner.Run(ctx); rerr != nil {
 		slog.Error("task failed", "error", rerr)
--- a/cmd/leoflow-server/main.go
+++ b/cmd/leoflow-server/main.go
@@ -697,7 +697,7 @@ func runGatedTicker(ctx context.Context, name string, ticks <-chan time.Time, le
 	}
 }
 
-func startReconciler(ctx context.Context, cs kubernetes.Interface, namespace string, reporter executor.FailureReporter, leading func() bool, logger *slog.Logger) {
+func startReconciler(ctx context.Context, cs kubernetes.Interface, namespace string, reporter executor.OutcomeReporter, leading func() bool, logger *slog.Logger) {
 	rec := executor.NewReconciler(cs, namespace, reporter)
 	startGatedTicker(ctx, "pod-reconcile", reconcileInterval, leading, logger, func() {
 		if err := rec.Reconcile(ctx); err != nil {
--- a/docs/adr/0052-durable-task-outcome.md
+++ b/docs/adr/0052-durable-task-outcome.md
@@ -2,7 +2,7 @@
 
 **Status:** Proposed
 **Date:** 2026-08-13
-**Relates:** ADR 0051 (separate the orchestration and execution state machines — this ADR is its Phase 2), ADR 0031 (scheduler reconciliation loop + two-layer reaping), ADR 0015 (Kubernetes-only container execution), ADR 0004 (thin agent), ADR 0002 (pod-per-task)
+**Relates:** ADR 0051 (separate the orchestration and execution state machines — this ADR is its Phase 2), ADR 0031 (scheduler reconciliation loop + two-layer reaping), ADR 0015 (Kubernetes-only container execution), ADR 0004 (thin agent), ADR 0002 (pod-per-task — an assumption this ADR is careful not to deepen)
 **Issues:** #543 (agent exit code conflates task outcome with report delivery); follow-up to #542 (in-process report retry)
 
 ## Context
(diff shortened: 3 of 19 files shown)


neochaotic:
## Reformulated: two-schools framing + design review + N:1 pod future

Rewrote the ADR after a design review of the first draft and primary-source research on Temporal, Argo, and Airflow-K8s.

**Correctness framing (the big change).** Named the two schools the industry splits on:
- **School A** — recover the lost outcome from a per-pod artifact (best-effort, node-loss blind).
- **School B** — the outcome doesn't exist until the *orchestrator* durably records it; on ambiguity, **re-drive + idempotency** (Temporal/Borg; correct under any failure).

leoflow's correctness **floor is already School B** (ADR 0031 reconcile/reap + ADR 0051 Phase 1 infra re-place). This ADR's termination-message record is positioned as a **School-A optimization on top** — it spares needless re-runs of costly/non-idempotent tasks when the kubelet survives, **not** the correctness mechanism. Absent the record, tasks still settle safely by re-drive.

**Design-review fixes folded in:**
1. `terminationMessagePolicy`/path is now an **explicit, tested prerequisite** in `BuildPod` (today it works only by the implicit K8s default).
2. Corrected the idempotency claim — the reconciler path (`FailTaskInstanceIfActive`) has **no `try_number` guard**, unlike `ReportTaskResult`; settling *success* through it could clobber a live retry. Specified a new **try_number-guarded settle** for both success and failure.
3. **Narrowed coverage** — only surviving-kubelet kills (OOM, node-pressure eviction, ActiveDeadline) are recovered; node-loss successes degrade to phase→failure→idempotent re-drive.
4. **Mandated write ordering** — success record written only after every pre-report push (return value, links, XCom) is accepted, so a recovered success can't have missing downstream data.
5. **`reschedule_at`** now carried in the reschedule record.

**Alternatives** now compare Temporal/Cadence, Argo (emissary exitcode + wait sidecar → same node-loss blind spot, `Error 'pod deleted'`), and Airflow-K8s zombie handling (heartbeat timeout → failed/retry, never recovers a success — the anti-pattern we beat).

**Product-direction note:** a task stays atomic to **one** pod, but that pod may be **dedicated** (exits at task end → termination message works) or **shared** with sibling tasks (a warm worker that doesn't exit per task → needs an in-band per-attempt transport, i.e. Temporal's model). Orchestration consumes outcome via the `ExecutionStore` seam, never pod phase — so the shared-pod future adds a transport without changing the orchestration side.

`d94aefc`. Still **Proposed** — no code until this framing is agreed.

neochaotic:
## Step 2 (Core, no cluster) — implemented, TDD

Four commits landing the durable-outcome mechanism end to end (unit + integration green, lint 0, build OK on go1.26.6):

**`internal/taskoutcome`** — the versioned record contract (`{v,outcome,exit_code,reschedule_at}`), shared by writer (agent) and reader (reconciler). `Decode` is defensive: empty/garbage/log-tail/unknown-version/reschedule-without-time all return not-ok → reader falls back to pod phase.

**Agent (`internal/agent/runner.go`)** — writes the record to `TerminationLogPath` **before** the report gRPC, keyed off the reported state inside `report()` (so `fail()` **and** the timeout sink `failWithReason()` are both covered — the gap the review flagged) + `reportReschedule()`. Best-effort, no-op when the path is empty (Lite). Tests: 3 outcomes, timeout sink, durability when the terminal report itself fails, Lite no-path.

**Executor (`BuildPod`)** — pins `TerminationMessagePolicy=File` + path (the review's prerequisite), passes the path to the agent via `podEnv`; Lite unaffected.

**Reconciler + storage** — `classifyPod` returns a rich verdict trusting a decodable record over phase; `OutcomeReporter` (was `FailureReporter`) gains `SucceedTask`/`RescheduleTask`, all **try_number-guarded**. New `SucceedTaskInstanceIfActive` + `RescheduleTaskInstanceByIDIfActive` + a try_number guard on `FailTaskInstanceIfActive` (id-only would let a stale reconciler clobber a live retry). Missing try-number label → skip settle, never GC.

**Integration (real Postgres)** proves the guard: stale attempt's success/failure does NOT land on the live retry; a recovered success does NOT resurrect a terminal row; and the headline — a running TI recovered to success (the zombie-task false-negative, closed).

Remaining: **Step 3** (k3d E2E — kill mid-report) runs on a cluster, separate effort. Still **Proposed** — not merging without your word.

neochaotic:
## Adversarial review of Step 2 — 1 blocker + 2 gaps, fixed (`55a71b0`)

A session-end adversarial review read all four feature commits + the surrounding pod/reaper/redispatch machinery. Findings and fixes:

**B1 (BLOCKING) — reschedule sensor flap.** The try_number guard protects fail/success because retry *bumps* try_number. But reschedule redispatch **reuses the same try_number** (`RedispatchRescheduledTaskInstance`, no attempt consumed, #380), so a lingering poke pod could re-park the re-dispatched *live* attempt with an already-elapsed `reschedule_at` → sensor flaps. `pod_name` can't disambiguate (column is never populated), and the scheduler has no pod manager to delete at redispatch. **Fix:** the reconciler collects a reschedule pod immediately after settling it (clean exit, nothing to inspect) — gone within one tick, well before the next redispatch (minutes). Bounded even for pathological sub-cadence pokes (≤1 self-correcting re-park).

**I3 (IMPORTANT) — pod leak.** A pod missing the try-number label was skipped for settle AND had GC deferred *forever* (label never appears). **Fix:** skip the settle but let it age out; the heartbeat reaper is the TI backstop.

**I2 (IMPORTANT) — reaper race, documented.** Success recovery (~30s) loses to the heartbeat reaper (~90s) if the reaper fails the still-running TI first → the recovered success is dropped and degrades to safe re-drive. Cadence makes the reconciler win the common case; documented as best-effort (School-A optimization on the School-B floor).

**Minors:** added the negative write-side tests the design leaned on (push-failure→FAILED record; RUNNING→no record); tightened two comments. M4 (exit_code/duration columns not populated on recovered outcomes) left as acknowledged observability-only.

**Confirmed correct by the review:** write ordering (success only after pushes; both failure sinks incl. timeout; record before every terminal RPC), classifyPod precedence/filtering, the try_number guard for fail/success (proven vs real Postgres), interface-change ripple (no stragglers), encoding/caps/tz, idempotency for fail/success.

Unit + integration green, lint 0, govulncheck 0 called. Step 3 (k3d E2E: kill mid-report + the flap-prevention end-to-end) remains — cluster work. Still **Proposed**.

PR #618, opened 2026-08-14
test(e2e): durable-outcome recovery chaos scenario (ADR 0052 Phase 2 Step 3)

neochaotic:
Closes the ADR 0052 loop: a full-cluster chaos scenario proving the reconciler recovers a task's success from the durable outcome record when the pod is killed mid-report.

## What

**Scenario C** in `test/e2e/chaos-runtime.sh`: a `recoverdag` whose `quick` task succeeds, but whose agent is told (via the task env `LEOFLOW_CHAOS_DIE_BEFORE_REPORT=TASK_STATE_SUCCESS`) to exit 137 **after** writing the durable outcome record and **before** delivering the report — a faithful OOM/eviction-mid-report. The pod ends Failed with no report; the scenario asserts the run **recovers to SUCCESS** from the record (not fail/retry), and that recovery is a **settle, not a retry** (try_number stays 1).

**The seam:** `Runner.BeforeReport` fires after the record is on disk and before the report RPC; the agent binary wires it from the env var to `os.Exit(137)`. The library core stays free of `os.Exit`/env. Never set in production.

## Verified here
- Unit: `TestRunnerBeforeReportHookFiresAfterRecordBeforeReport` locks the ordering (record on disk, terminal report not yet sent when the hook fires).
- The `recoverdag` + `leoflow.yaml` compile and the chaos env binds to the `quick` task in `dag.json`.
- `bash -n` + shellcheck clean; full build + agent unit + golangci-lint 0.
- (Step 2's reconciler recovery + attempt guard are already unit + integration tested on main.)

## Needs a cluster run
The runtime chaos harness is **not CI-gated** (destructive, slow) — run it on the Lima Linux VM:
```bash
test/e2e/chaos-runtime.sh
```
This PR's CI validates the code (agent seam, build, lint, all unit + integration + the CI-gated k3d E2Es), but **not** Scenario C itself. The final green for Step 3 is a Lima run of the harness.

### DIFF
--- a/cmd/leoflow-agent/main.go
+++ b/cmd/leoflow-agent/main.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/neochaotic/leoflow/internal/agent"
 	"github.com/neochaotic/leoflow/internal/version"
+	agentv1 "github.com/neochaotic/leoflow/proto/agent/v1"
 )
 
 // usage is printed for `--help`. leoflow-agent takes no positional args; it is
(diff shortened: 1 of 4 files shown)


neochaotic:
Rebased onto main (Wave 0+1). **Local smoke on a healthy Lima/k3d VM: chaos-runtime.sh PASSED green — all three scenarios.** Scenario C (ADR 0052) genuinely fired this time (the hardened base-image guard + fail-closed assertion did their job): the pod ended `Error` with the outcome record already written, and the reconciler recovered the run to SUCCESS from the durable record despite the Failed pod and lost report, settling try_number=1 with no wasteful retry. Also validated the operator E2E (test/e2e/e2e.sh) green on main. This closes the clean-run gate this PR was held on.

Issue #852, opened 2026-09-01
compile: synthesized Dockerfile installs deps as non-root → dbt (and any CLI dep) off PATH


neochaotic:
Confirmed in code. A dbt DAG built via `leoflow compile --build` (the synthesized Dockerfile path — no user Dockerfile) produces an image where `dbt` is not on PATH, so the task fails `dbt: command not found`. Any dependency that ships a console-script entrypoint has the same problem.

## Root cause
- The runtime base ends with `USER 65532:65532` (`runtime/Dockerfile:55`) and sets no `PATH` that includes `~/.local/bin`.
- `generatedDockerfile` (`internal/cli/compile_build.go`) emits `RUN apt-get …` and `RUN pip install <deps>` with **no `USER root`**, so they run as the inherited non-root `65532`:
  - `pip install` as non-root falls back to a `--user` install → console scripts land in `/home/leoflow/.local/bin`, which is **not on PATH** → `dbt` not found.
  - `apt-get install` as non-root fails outright (needs root), so `system_packages` in a synthesized build is also broken.
- Pure-import Python DAGs escape (the `--user` site is on `sys.path`), so only CLI-invoked deps (dbt) and `system_packages` are affected.

## Why the e2e did not catch it
`test/e2e/dbt-connection-e2e.sh` builds a **hand-written** Dockerfile that does `USER root` before `pip install` (+ `chown -R 65532`), then drops back to `USER 65532`. So the e2e proves a Dockerfile shape real users don’t get from `compile --build`. The synthesized path is untested.

## Fix
In `generatedDockerfile`: switch to `USER root` for the apt/pip install layers (console scripts → `/usr/local/bin`, on PATH), then drop back to `USER 65532:65532` before the runtime, and `chown -R 65532:65532` the COPYed dbt project (dbt writes `target/` + `profiles.yml` at runtime as the non-root user). This mirrors the pattern the working e2e already uses. Keep the final `USER 65532` numeric for PodSecurity `runAsNonRoot`.

## Also
Change `dbt-connection-e2e.sh` (and/or add a case) to build via the **synthesized** Dockerfile (deps in `leoflow.yaml`, no hand-written Dockerfile) so this class is gated. Add a unit test on `generatedDockerfile` asserting `USER root` precedes the install and `USER 65532` ends it.

## Severity
[...]

PR #854, opened 2026-09-01
fix(compile): dbt image installs deps as root + writes artifacts to /tmp (#852)

neochaotic:
Field feedback (issue #852), refined fix. A dbt DAG built via `leoflow compile --build` (the synthesized Dockerfile) produced a broken image: `dbt: command not found`.

## Root cause
The synthesized Dockerfile ran `apt`/`pip` as the base image’s non-root `USER 65532`, so pip fell back to a `--user` install and dbt’s console script landed in `~/.local/bin` — **not on PATH**; `apt` failed outright. The e2e masked it by hand-writing a Dockerfile with `USER root`.

## Fix (refined — read-only project, no chown)
- `generatedDockerfile`: install deps as **root** (console scripts → `/usr/local/bin`, on PATH), then drop back to `USER 65532:65532`. The COPYed dbt project stays **root-owned / read-only** to the task.
- `runtime/Dockerfile`: point `DBT_PROFILES_DIR` / `DBT_TARGET_PATH` / `DBT_LOG_PATH` at an ephemeral `/tmp/leoflow/dbt` dir, so dbt writes its artifacts there — never the read-only project. Harmless for non-dbt tasks.

## Security posture
Build-time root is standard (apt/pip need it); the **runtime** container is non-root (`USER 65532`, PSA `runAsNonRoot`). The project being read-only is a *stronger* posture (a task cannot modify its own code), and routing dbt’s writes to `/tmp` lets a task pod run with `readOnlyRootFilesystem: true` + a `/tmp` emptyDir (follow-up).

## Tests
- `generatedDockerfile`: USER root precedes pip and USER 65532 is the last USER (runtime non-root); no chown (project read-only); no-deps build needs no root switch.
- Existing generatedDockerfile tests still green; `golangci-lint --new-from-rev origin/main` 0 issues.

## Follow-up
Refactor `dbt-connection-e2e.sh` to build via the **synthesized** Dockerfile (deps in leoflow.yaml, no hand-written Dockerfile) so this class is gated in CI; and flip the chart to `readOnlyRootFilesystem: true` with a `/tmp` emptyDir.

### DIFF
--- a/internal/cli/compile_build.go
+++ b/internal/cli/compile_build.go
@@ -67,6 +67,15 @@ func generatedDockerfile(cfg *domain.LeoflowConfig, dagSource string) (string, e
 	}
 	var b strings.Builder
 	fmt.Fprintf(&b, "FROM %s\n", resolveBaseImage(cfg))
+	// apt requires root, and pip as the base's non-root USER (65532) falls back to a
+	// `--user` install whose console scripts (e.g. `dbt`) land in ~/.local/bin, which
+	// is not on PATH — so a synthesized dbt image would fail `dbt: command not found`.
+	// Install as root (system site → /usr/local/bin on PATH), then drop back to the
+	// non-root runtime USER before the source COPY (#852).
+	rootForInstall := len(cfg.SystemPackages) > 0 || len(deps) > 0
+	if rootForInstall {
+		b.WriteString("USER root\n")
+	}
 	if len(cfg.SystemPackages) > 0 {
 		// Single RUN so the apt cache cleanup stays in the same layer as the install.
 		fmt.Fprintf(&b, "RUN apt-get update && apt-get install -y --no-install-recommends %s "+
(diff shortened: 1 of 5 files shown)


neochaotic:
Security review came back: **net security improvement, not a regression** (runtime stays non-root `USER 65532`, read-only root-owned project, `/tmp` single-tenant per pod, ENV neutral + enables `readOnlyRootFilesystem`). Two follow-ups from it:

- **Applied here:** `profiles.yml` is now written **0600** (owner-only) — it carries warehouse credentials (commit `harden(runtime)…`).
- **Concern #3 (baked manifest vs DBT_TARGET_PATH) — verified, no regression.** Ran `dbt parse` on a **read-only** project with `DBT_TARGET_PATH=/tmp`: it succeeds, writing manifest/artifacts to `/tmp`, never the project (`Unable to do partial parsing … Starting full parse.`). `dbt run` always re-parses to the target path at runtime — the baked `target/manifest.json` is a compile-time artifact leoflow uses to render the DAG, not the runtime source — so pointing the target at `/tmp` changes nothing functionally.

**Deferred follow-ups (tracked in #852):** (a) make the `/tmp` emptyDir tmpfs (`Medium: Memory`) + a `SizeLimit` when flipping `taskPodSecurity.readOnlyRootFilesystem: true`, so warehouse secrets stay in RAM; (b) refactor `dbt-connection-e2e.sh` to build via the synthesized Dockerfile so this class is gated in CI.

Issue #1006, opened 2026-09-08
docs/code housekeeping from the dbt authoring review


neochaotic:
Three small drifts found while reviewing the dbt authoring path, grouped because each is a one-liner.

1. **Granularity comment claims a value the schema does not have.** `internal/domain/config.go:88-89` says "node, level, folder, or **tag** (ADR 0042 §5)", but the authoring schema's enum is `["node","level","folder"]` and its own description notes tag grouping is #398. The website is correct; the Go comment is the drifted one — and it is the kind of comment the next doc gets written from.
2. **The dbt reference table omits `schema`.** `website/content/author-dags/dbt.md:473-481` covers both `dbt:` and `dbt_groups:` but not the target-schema override (`internal/domain/config.go:103-105`), which exists in both blocks.
3. **`examples/README.md:3`** says "Twenty reference projects" against 23 directories.

Issue #1005, opened 2026-09-08
Verify whether task pods reuse the build-time dbt parse, or re-parse every run


neochaotic:
`website/content/author-dags/dbt.md:361-362` and `:369-370` claim runtime pods reuse the parse done at build time. Strongly indicated to be **false**, but not proven — it wants one measurement.

Evidence it is false: the base image sets `DBT_TARGET_PATH=/tmp/leoflow/dbt/target` (`runtime/Dockerfile:71-73`) and the Lite subprocess executor points it at a per-task scratch dir (`internal/executor/subprocess.go:136`). dbt reads `partial_parse.msgpack` from its target path, and nothing in the tree copies the baked `target/` there — `grep partial_parse` outside comments returns zero hits.

If it is false it is the most expensive error on that page: the pageinfo claim at `:20-21` ("no per-run re-parse") and the Cosmos comparison row at `:488` both rest on it.

**The measurement:** run one dbt task in a pod with `dbt --debug` and read whether it reports a full parse or a partial one. Twenty minutes on the validation cluster.

neochaotic:
Answered, locally, in about a second — no cluster needed.

Ran the exact command the RC runbook §3a now prescribes, against the DAG image the mixed-mode e2e builds:

```
$ docker run --rm --entrypoint dbt leoflow-dbt-mix-dag:dev --debug parse \
    --project-dir transform --profiles-dir transform
...
target_path: /tmp/leoflow/dbt/target
Unable to do partial parsing because saved manifest not found. Starting full parse.
Wrote artifact WritableManifest to /tmp/leoflow/dbt/target/manifest.json
Command `dbt parse` succeeded after 1.17 seconds
```

**A task pod does a full parse every run.** dbt says so in as many words, and the reason is structural: `DBT_TARGET_PATH` is `/tmp/leoflow/dbt/target` (set in `runtime/Dockerfile`), which is empty in a fresh container, so `partial_parse.msgpack` cannot be there. The build-time `dbt parse` writes its `target/` inside the project directory, and nothing copies one to the other — I grepped every reference to `DBT_TARGET_PATH` in the tree to confirm there is no such copy.

So the removed docs claim ("runtime pods reuse the build-time parse") was correctly removed in #1009, and should stay out.

[...]

Issue #1004, opened 2026-09-08
docs: dbt_group has no reference page


neochaotic:
`dbt_group()` is the primitive the whole authoring model rests on, and it is absent from the Python reference.

`website/scripts/gen-python.sh:27-31` passes only `leoflow_runtime*` to pdoc, so the `leoflow` authoring package (`runtime/python/leoflow/__init__.py`, added in #991) is never rendered. `website/content/reference/python-api.md:33-37` does not name it either.

Add the module to the pdoc invocation and name `dbt_group` in the prose.

Issue #1003, opened 2026-09-08
examples: no dbt example, and the gap is what let #17 and #20 ship


neochaotic:
`examples/` has 23 projects and **zero** `dbt_project.yml` — verified with `find`. So for dbt the doc snippets are the only artifact, and they are broken.

Add `examples/dbt_hybrid/`, and make it close the coverage gap rather than just illustrate:

- a **`PythonOperator` / `@task`**, not a `BashOperator`. That is what forces the runtime's per-task re-import of `dag.py`, which is the mechanism that made `from leoflow import dbt_group` explode in the pod (#17). `test/e2e/dbt-mixing-e2e.sh` used Bash tasks and structurally could not catch it.
- **no hand-written Dockerfile**, so the generated one — which had zero `DbtGroups` references (#20) — is what actually runs.
- `dependencies: ["dbt-duckdb"]` and **no `connection:`**, so it runs on Lite with no warehouse setup. A dbt example that needs a Postgres is an example nobody runs.
- a committed `target/manifest.json` plus a `dbt.manifest:` pin, so it works for anyone without dbt on PATH.

Also: `examples/README.md:3` says "Twenty reference projects" against 23 directories.

PR #1009, opened 2026-09-09
docs: make the dbt page teach a DAG, and drop four false claims

neochaotic:
The docs are why a field team spent their effort on the wrong shape. The steering is upstream of every code defect we fixed this week — and the belief did **not** form on the dbt page. By the time a reader got there, they had been sorted two clicks earlier.

## Where the wrong belief formed

**Homepage.** "Author" and "Run dbt as DAGs" sat side by side as *peers* in the feature grid. Someone arriving with a dbt project reads "Author" as *not me*. Now: **"dbt in your DAGs"**, leading with `dbt_group()` in a `dag.py`.

**Section index.** dbt was the `lf-card--hero` with a `Highlight` badge, *above* "DAG authoring" — while the sidebar already had the correct order, so the two navigations disagreed and the visual one won. The hero is back on DAG authoring.

**The page.** It asked *"There are two ways — pick the one that fits your DAG"*, with the pure-dbt path as a hero badged **"Simplest"** and the hybrid as a plain card described as "the Cosmos `DbtTaskGroup` capability" — one in our own vocabulary, one as a port of someone else's class. The chooser is deleted. The page opens on a `dag.py`.

## What "Simplest" cost

It was true about the first twenty minutes and false about everything after — and it graded the wrong axis. The reader is not choosing between two efforts; they are choosing between two ceilings with a migration in between.

The shortcut section now states its exit cost: no sensors, no provider operators, no Python or Bash, and nowhere to declare `start_date`, `end_date`, `catchup`, `max_active_runs`, `max_active_tasks` or DAG-level `params`/`connections`/`variables`. Plus the sentence that would have saved the field team: **every `task_id` gains the group prefix**, so leaving the shortcut is a migration, not an edit.

## Four factual errors

| Claim | Reality |
|---|---|
| Both flagship snippets | omitted the dbt adapter from `dependencies:` — copying either built an image whose first dbt task dies |
| "`dbt parse` runs at image-build time" | zero occurrences in the Dockerfile generator; it runs on your machine in `leoflow compile`, both editions |
| "runtime pods reuse the build-time parse" | strongly indicated false and unverified — removed rather than left standing (#1005) |
| ADR 0042's config fence | teaches `groups:` and a six-value `granularity` enum against a shipped three, in a doc marked "shipped in v0.1.1" |

The third is removed rather than corrected because I cannot prove either way yet; it returns if the measurement in #1005 supports it.

## Deletions

The page had accreted — 491 lines. Removed: the chooser, three "vs Cosmos" asides the comparison table already covered in full, 39 lines of CI self-reporting (replaced by an adapters-and-auth table an author can act on), and a Slim CI subsection explaining a feature we do not have. The pre-build-the-manifest block moves out of the shortcut section — it applies to both shapes — and becomes its own section, which also fixes the anchor pointing at it.

## Verification

`hugo --gc --minify` builds clean, every internal anchor resolves, no chooser card remains, all `scripts/check-*.sh` green.

## Not in this PR

The code items the UX review surfaced are filed: #1001 (`dbt:` + `dag.py` silently discards the Python), #1002 (`leoflow init --dbt`), #1003 (`examples/dbt_hybrid/`), #1004 (`dbt_group` has no reference page), #1005, #1006.

### DIFF
--- a/website/content/_index.md
+++ b/website/content/_index.md
@@ -34,10 +34,10 @@ compile it to an immutable image. Native **map-reduce** for ML/AI as a Python li
 comprehension.
 {{% /blocks/feature %}}
 
-{{% blocks/feature icon="fa-solid fa-cubes-stacked" title="Run dbt as DAGs" url="/author-dags/dbt/" url_text="dbt projects as DAGs" %}}
-Render a **dbt** project into native model-level tasks — pod-per-task against your
-warehouse, no Airflow and no Cosmos at runtime. One `granularity` knob trades pods
-for speed.
+{{% blocks/feature icon="fa-solid fa-cubes-stacked" title="dbt in your DAGs" url="/author-dags/dbt/" url_text="Put dbt in a DAG" %}}
+Drop `dbt_group()` into your `dag.py` and your **dbt** project's models become tasks
+in the same graph — pod-per-model against your warehouse, no Airflow and no Cosmos
+at runtime. Your `ref()` graph is the shape; you never write the dependencies.
 {{% /blocks/feature %}}
 
 {{% blocks/feature icon="fa-solid fa-plug" title="Connect" url="/connections/" url_text="Browse connectors" %}}
--- a/website/content/author-dags/_index.md
+++ b/website/content/author-dags/_index.md
@@ -18,18 +18,18 @@ the project layout and compile model, then reach for the guide that matches your
 task below.
 
 <div class="lf-cards">
-  <a class="lf-card lf-card--hero" href="/author-dags/dbt/">
-    <span class="lf-card__badge">Highlight</span>
-    <span class="lf-card__icon"><i class="fa-solid fa-cubes-stacked"></i></span>
-    <span class="lf-card__title">dbt projects as DAGs</span>
-    <span class="lf-card__desc">Render a dbt project into native model-level tasks — pod-per-task, no Airflow and no Cosmos at runtime.</span>
-    <span class="lf-card__more">Run dbt as a DAG →</span>
-  </a>
-  <a class="lf-card" href="/author-dags/dag-authoring/">
+  <a class="lf-card lf-card--hero" href="/author-dags/dag-authoring/">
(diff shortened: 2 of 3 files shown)


neochaotic:
Docs review came back with two blocking items, both about the honesty of the limitations list. Both fixed.

**1. The exit-cost box credited `dag.py` with two things it does not do.** It listed six DAG-level settings the shortcut cannot declare — implying the DAG-based path declares all six. It declares four. `compiler.py` reads exactly eight attributes off the DAG object:

```
$ grep -oE 'getattr\(dag,\s*"[a-z_]+"' parser/leoflow_parser/compiler.py
tags description max_active_runs catchup default_args timetable schedule start_date params
```

No `end_date`, no `max_active_tasks`. And `grep -c 'end_date\|max_active_tasks' docs/api/leoflow-yaml-schema.json` → **0**. They are a product-wide authoring gap (#797), not a cost of taking the shortcut. Listing them as one overstated the very path this page steers people toward — which is the class of error the rewrite exists to remove, so I am glad it was caught here rather than by a reader.

**2. Added the #1001 trap.** A project carrying both a top-level `dbt:` block and a `dag.py` compiles green and **silently discards the Python** — the routing happens before the parser is consulted. This is the exact incident that motivated the rewrite, it is filed, and the migration paragraph gave the right order without saying what getting it wrong costs. It now says *delete `dbt:` first*.

Two smaller ones in the same push:

- `alerts` and `staging` were labelled per-task. They are DAG-wide — the `tasks:` sub-schema is `additionalProperties: false` with neither key, so the old wording described a config that gets rejected.
- The resumable-fused-retry line now cites #569 and reads as **Planned** rather than a vague "future enhancement". The issue existed; it just was not linked.

`hugo --gc --minify` clean.

One review observation I am recording rather than acting on: the compressed adapter section lost the Databricks live-verification narrative (view + Delta table materialized, cold start observed and non-blocking). The substance — which adapters are contract-tested vs live-tested — survives in the table. I think the trade is right for a page this long, but flagging it in case you want the sentence back.

Issue #993, opened 2026-09-08
compile without --build bakes an absolute host path into dbt task entrypoints


neochaotic:
`dbtProjectDir(dir, project, local=true)` returns `filepath.Abs(dir/project)`, and `local` is `!o.build`. So `leoflow compile` without `--build`, and `leoflow deploy --skip-build`, write the developer's absolute host path into every dbt task's entrypoint — and that dag.json is what gets registered and executed in pods.

Measured on a dbt-only project:

```
entrypoint: python -m leoflow_runtime --dbt-default-duckdb analytics /var/folders/.../analytics/leoflow_local.duckdb && dbt run --select mart --project-dir /var/folders/.../analytics
```

Same symptom as #20 — "project directory does not exist", seconds after pod start, after a green compile — through a different door, and the more likely one for a CI/prebuilt-image workflow. Affects both the top-level `dbt:` block and `dbt_groups`.

Found while reviewing #992. Not fixed there.

Issue #729, opened 2026-08-24
fix(setup): managed Postgres install is not idempotent across binary upgrades


alissonrosa-lang:
## Summary

`leoflow lite --postgres managed` fails hard after a binary upgrade, because the managed-Postgres extraction re-creates symlinks that already exist on disk.

## Reproduction

1. Run `leoflow lite --postgres managed` on release *A*. Managed Postgres installs under `~/.leoflow/postgres` and works.
2. Upgrade the `leoflow` binary to release *B*.
3. Run `leoflow lite --postgres managed` again:

```
downloading relocatable PostgreSQL 16.13.0 for darwin/arm64...
error: installing managed Postgres: symlink libecpg.6.dylib
       /root/.leoflow/postgres/lib/libecpg.dylib: file exists
```

Hit on darwin/arm64 going from v0.2.0 to a v0.4.0-rc build, against a `~/.leoflow` that release A had created successfully.

## Root cause

`internal/setup/install.go:207` ends the symlink branch of the tar extractor with:

```go
return os.Symlink(hdr.Linkname, target)
```

`os.Symlink` fails with `EEXIST` if the path exists. The regular-file branch overwrites happily, so only symlinked entries break — which is why a partial re-extract is enough to wedge it. The path-containment checks just above are careful and correct; it is only the create that is not idempotent.

## Follow-on failure

After clearing `~/.leoflow/postgres` to get past the above, the next boot failed differently:

```
error: provisioning venv for project "…": creating dev venv for "demo" with /usr/bin/python3
       (the managed CPython bundles venv; a system python3 may need its python3-venv package): exit status 1
```

It had fallen back to the system interpreter rather than the managed CPython under `~/.leoflow/python`. A full `rm -rf ~/.leoflow` + `leoflow setup` recovered. So a half-upgraded install does not self-heal, and the failure surfaces at a component unrelated to the cause — there is a `lite-selfheal` gate for the boot path, but it does not appear to cover "existing install, newer binary".

## Suggested fix

Remove the target before creating the link (`os.Remove` ignoring `IsNotExist`, then `os.Symlink`), or `RemoveAll` the destination directory before extracting when a re-install is intended. Given `leoflow setup` documents that "re-running is safe", an upgrade-path case in the self-heal gate — install with the previous release's binary, then boot with the current one — would be worth more than the one-line fix.

alissonrosa-lang:
[...]

Issue #722, opened 2026-08-24
fix(agentrpc): secret-scope warning audit write always fails — tenant UUID passed to a name lookup


alissonrosa-lang:
## Summary

Every `secret.scope_warning` audit write fails. The observe-mode trail that ADR 0055 relies on before flipping `auth.secret_scoping` to `enforce` records nothing.

## Root cause

`internal/agentrpc/secrets.go:338` passes `id.TenantID` — a **UUID** — as the `tenant` argument:

```go
s.secretAudit.RecordSecretScopeWarning(ctx, id.TenantID, id.DagID, id.RunID, id.TaskID, kind, declaredCount, deliveredCount)
```

`RecordSecretScopeWarning` (`internal/storage/repository.go:723-724`) resolves that argument by **name**:

```go
tid, err := r.tenantID(ctx, tenant)   // -> GetTenantByName(ctx, name)  (repository.go:81-87)
```

A UUID never matches a tenant *name*, so `mapNotFound` turns it into `resource not found` and the audit row is never written. The log warning itself is emitted first (`secrets.go:332`), so the failure is easy to miss.

## Reproduction

Lite, subprocess executor, defaults (`secret_scoping: permissive`). Define 3 connections + 2 variables; register a DAG declaring only one of the connections; trigger it.

```
WARN task received the full secret vault but declared a narrower set; under secret_scoping: enforce
     it would receive only its declared set ... dag=demo task=sonda kind=connections declared=1 delivered=3
WARN recording secret-scope warning audit event ti=... kind=connections error="resource not found"
```

`GET /api/v2/eventLogs` returns the `dagrun.manual.trigger` events but no `secret.scope_warning` — 1 warning, 1 audit failure, 100% loss.

## Impact

The `permissive`/`observe` → `enforce` rollout is the documented path to closing the full-vault delivery (ADR 0055 D9). Its observation period currently produces log lines only, so an operator auditing readiness through the audit log sees a clean trail and concludes there is nothing to migrate.

## Suggested fix

Either resolve the tenant by id in `RecordSecretScopeWarning`, or have it accept the already-resolved `pgtype.UUID`. Worth checking the other `secretAudit` call sites for the same name/UUID mismatch, and a test asserting the audit row actually lands (the current unit tests assert the warning, not the write).

Issue #994, opened 2026-09-08
A dbt group with no connection cannot resolve a project-baked profiles.yml


neochaotic:
The base image sets `DBT_PROFILES_DIR=/tmp/leoflow/dbt`, which does not exist. A dbt task with no `connection:` takes `decorateCommands`' default branch (`internal/dbt/render.go:196` — "the image's baked profiles.yml") and gets no profile prefix, so dbt never reads the `profiles.yml` the COPY brought in.

Measured in a built image:

```
dbt parse --project-dir ./transform
  -> Error: Invalid value for '--profiles-dir': Path '/tmp/leoflow/dbt' does not exist
after mkdir -p /tmp/leoflow/dbt
  -> Runtime Error: Could not find profile named 'transform'
```

Pre-existing and shared by the top-level `dbt:` block and `dbt_groups`.

PR #1012, opened 2026-09-09
docs(release): point the RC runbook at v0.4.5, and stop it certifying the wrong cloud

neochaotic:
`test/release/rc-cluster-validation.md` says to refresh §3 per RC and to be reusable "per RC and per cloud". It was neither: §3 still described the rc.3 tranche (#722–#729), and the run header hardcoded **AWS EKS** and `v0.4.0-rc.3`. Run as written against this release, it would have validated the previous one's features.

## §3 → v0.4.5

The five GA blockers this release closed, in the file's own notation — ✔ for what unit or e2e already proves, ★ for what only a pod shows.

That split carries more weight than usual here, because #1008 makes `dbt-mixing-e2e.sh` exercise the generated path end to end. Most rows are now ✔, and the cluster's remaining job is narrow and specific:

| | why a cluster |
|---|---|
| **#993** | needs two machines — compile on one, `deploy --skip-build` from another, then read the pod's args for a host path. The e2e structurally cannot reach it |
| **#852** | `readOnlyRootFilesystem` with a real `/tmp` emptyDir, and `runAsNonRoot` with no `runAsUser` so the **kubelet** resolves the image's numeric USER — a different code path from `docker run`, where a symbolic `USER nonroot` passes and in a pod does not |
| **#1005** | one `dbt --debug` in a task pod settles whether the build-time parse is reused. Twenty minutes, and it decides whether a claim returns to the docs or stays out |

## The cloud half

§5 already lists four EKS↔GKE deltas "so a GKE pass doesn't give false confidence" — but §6 offered only PASS/FAIL, leaving an operator nowhere to put *"this cloud cannot settle it"*. §6 now says to record **NOT VERIFIED (cloud)** for those rows, and §3 ends with which of this tranche's rows that applies to (NetworkPolicy enforcement; nothing here depends on IRSA, RWX or ingress).

A blank row reads as "not run". A PASS on the wrong cloud reads as proof. That second one is the failure §5 exists to prevent, and the results table was quietly inviting it.

## Header

Tag and cloud are placeholders now, and the CNI row says to note it because §5 depends on it.

The one surviving `v0.4.0-rc.3` reference is a historical note about that release shipping with a stale `Chart.yaml`. It is about the past and stays.

Docs only — no code, no test changes.

### DIFF
--- a/test/release/rc-cluster-validation.md
+++ b/test/release/rc-cluster-validation.md
@@ -180,7 +180,8 @@ kubectl run img-probe -n <taskNamespace> --rm -i --restart=Never \
 
   ```bash
   API=<control-plane base URL>          # e.g. http://localhost:8080 via port-forward
-  TOKEN=$(leoflow auth token)           # or the JWT from §2's login
+  TOKEN=$(leoflow auth create-token --server "$API" \
+                  --username <u> --password <p>)
   DAG=<the hybrid DAG from #17>         # NOT the §2 smoke DAG — it has no dbt tasks,
                                         # so it passes this vacuously
   curl -fsS -H "Authorization: Bearer $TOKEN" "$API/api/v2/dags/$DAG/spec" \
@@ -205,7 +206,7 @@ kubectl run img-probe -n <taskNamespace> --rm -i --restart=Never \
   succeeding. `decorateCommands` appends `--profiles-dir` only in the
   no-connection branch, so the fix sits one `switch` arm from the managed-secret
   path — the arm that also crosses ADR 0060 external secrets. Run one.
-- **#852 read-only project, for real ★** — the half no test covers.
+- **#852 read-only project, for real** — the half no test covers.
 
   ```bash
   kubectl label ns <taskNamespace> pod-security.kubernetes.io/enforce=restricted
@@ -222,15 +223,20 @@ kubectl run img-probe -n <taskNamespace> --rm -i --restart=Never \
   must resolve the image's numeric USER. That half is **already green three ways**
   (every k3d e2e pod, plus the image-level UID assertion in the e2e); it is not
   what makes this row cluster-only.
-  Run the PSA-`restricted` admission and the `LimitRange` ephemeral-storage
-  default here too — but note **k3d enforces both** (PodSecurity admission is
-  built in and on since 1.25; LimitRange is core), so a green there is not cloud
-  evidence and these belong in a k3d e2e rather than on cluster minutes.
-  What is genuinely cloud-only: **GKE Autopilot** mutating securityContext and
-  injecting resource requests, and **node-level ephemeral-storage eviction** under
-  a real kubelet with real disk pressure — the executor creates that emptyDir with
-  **no `sizeLimit`**, so nothing in our code bounds `/tmp/leoflow/dbt/target` on a
-  large manifest and the node is the only thing that will.
+  **This row has no cloud-only half, and three review rounds failed to find one.**
+  Each proposed anchor collapsed on inspection: the kubelet resolving the image's
+  numeric USER is already green three ways; PSA `restricted` admission and
+  `LimitRange` are in-tree apiserver features k3d enforces identically; node-level
+  ephemeral-storage eviction runs on a real kubelet on k3d too. GKE Autopilot
+  would be a genuine delta but the RC target is EKS, where it does not exist.
+  So: **run it on whatever cluster you have and record the result, but do not
+  spend cluster time you would not otherwise spend on it.** The right home for
+  this is a k3d e2e setting
+  `LEOFLOW_EXECUTOR_DEFAULTS_READ_ONLY_TASK_ROOT_FILESYSTEM=true` against a
+  `restricted`-labelled namespace, filed as #1016. One thing worth noting while
+  you are here, because nothing in our code bounds it: the executor creates that
+  emptyDir with **no `sizeLimit`**, so a namespace `LimitRange` or the node's own
+  eviction threshold is the only limit on `/tmp/leoflow/dbt/target`.
 - **#1005 does a task pod reuse the build-time parse?** — **run this locally, not
   on the cluster.** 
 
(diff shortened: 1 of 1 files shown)


neochaotic:
Round-2 review found a row that could not fail — pushed a fix.

**The blocker:** the #993 row read the entrypoint from `pod.spec.containers[0].args`. `BuildPod` (`internal/executor/kubernetes.go:113-131`) sets neither `command` nor `args` — the entrypoint reaches the task over gRPC (`internal/agentrpc/server.go` → `internal/agent/runner.go` `BuildCommand`), and `grep -rn Entrypoint internal/executor/` returns exactly one line, the struct field. So that jsonpath returns empty on a fixed build and on a broken one. An operator would have marked PASS having proven nothing, on the riskiest fix in the tranche. A gate that always passes is worse than no gate.

Replaced with a bare `leoflow compile` + the `/spec` entrypoint assertion (`internal/api/ui_tasks.go:245`), which does distinguish the two states. Two machines were never needed either — `dbtProjectDir` only absolutizes under `local`, and `dev.go:754` is the only production site that sets it.

Also fixed:

- `kubectl exec` into a task pod is not runnable (`restartPolicy: Never`, seconds of life) → throwaway `kubectl run --rm --restart=Never` from the DAG image.
- Marks corrected against the code: #17 and #994 lose their ★ (the `leoflow` package resolves from site-packages regardless of `PYTHONPATH`/UID; the `/tmp` emptyDir only exists when `readOnlyRootFilesystem` is on, which is off by default), #1005 becomes a local `docker run`, #993 gains its ✔. #852 keeps ★ but for the right reason — PSA `restricted` admission and an unsized emptyDir under a `LimitRange`, not the kubelet numeric-USER half that is already green three ways.
- §3 split into **3a** (per-RC tranche) / **3b** (standing). The old "refresh §3" instruction had swept away the two #800 blind-spot rows the CHANGELOG promises this release adds — with #800’s code half still open — plus four procedures (#722/#724/#727/#729) whose result rows survived without them. All restored.
- Added §4.5 (the PDB that now renders unasked on any `split`/`autoscaling` upgrade; ADR 0052 restart recovery) and the two uncovered arms: a dbt group **with** `connection:`, and `project: "."`.
- De-AWS’d §0/§1 and scoped the GKE caveat to the §4 rows it actually applies to.

Two review claims I did not apply, because they do not hold:

- `helm/leoflow/values-ha.yaml` does not exist in the tree, so the `karpenter.sh/do-not-disrupt` note went nowhere.
- The review suggested §4 already carried the #722/#724/#727/#729 procedures. It does not — `awk` over §4 finds none of those four. I had written that into §3b before checking, and removed it; the procedures are restored inline instead.

**Merge order:** #1008 first. Most ✔ marks in §3a rest on `dbt-mixing-e2e.sh` exercising the generated Dockerfile path, which lands there — §3a now says so explicitly rather than assuming it.

neochaotic:
Round 3. The repair had reproduced the defect it was repairing — worth stating plainly, because it is the third time today the same shape got through.

**The §4.5 ADR 0052 row I *added* could not fail.** It said: kill the control plane mid-task, confirm the task is not marked failed. That proves nothing — the agent’s terminal report retries for as long as the control plane is unreachable, so the ordinary report path settles the task to `success` without the recovery path running at all. The repo already refuses this exact false green (`test/e2e/chaos-runtime.sh:410-415`):

> if the pod ends `Completed` (exit 0) the fault-injection seam did NOT fire … the recovery path was never exercised — **a false green**

And the rationale I gave for the row ("k3d can stage the pod but not the eviction class of failure") was false: scenarios **C and D run on k3d**, C being ADR 0052 durable-outcome recovery. It was never cluster-only. The row now points at the script — and earns its place only because that script is **not in any workflow**, so an RC is when it actually gets run.

**The PDB step was inert.** I asked the operator to set `unhealthyPodEvictionPolicy=AlwaysAllow` "and confirm it renders". That is the chart default (`values.yaml:483`), so it always renders. Inverted to the falsifiable form: confirm the apiserver **kept** the field, because one older than 1.27 prunes it silently and hands you `IfHealthyBudget`, under which two unready replicas cannot be evicted and the drain hangs on an already-down control plane. Also added the precondition that makes the drain check honest — no default anti-affinity means both api replicas can land on one node, and a **correct** install would then read FAIL — and pointed the install at `helm/leoflow/examples/values-ha.yaml`, the HA profile we advertise and otherwise never validate.

**The #1005 command did not run.** `WORKDIR` is `/home/leoflow`, the project sits below it → bare `dbt parse` dies "Not a dbt project"; and `DBT_PROFILES_DIR` points at a `/tmp` path absent from a fresh container. Both flags now passed explicitly.

**#852’s ★ was anchored on the wrong things.** PSA `restricted` admission and `LimitRange` are in-tree apiserver features that k3d enforces identically — so they belong in a k3d e2e, not on cluster minutes. Re-anchored on GKE Autopilot’s spec mutation and node-level ephemeral-storage eviction, which need a real kubelet under real disk pressure.

Smaller, same push: the #993 cluster check pins the DAG (aimed at the §2 smoke DAG it passes vacuously) and defines `$API`/`$TOKEN`; its local half is now credited to `internal/cli/dbt_projectdir_test.go`, not to the e2e; §6’s #804 row asked for the object count §4.4 spends a paragraph warning against; #800’s positive half names an observation (probe task printing `AIRFLOW_CONN_*` **names**, never values); two §6 rows still said §3 instead of §3a; §0’s identity pointer named §4.3 (QoS/ingress) instead of §4.2 (IRSA).

Two review items I did **not** apply:

- **M1 is stale, not wrong.** The review checked `f65fcea` and found `-e PYTHONPATH=` only in this document. #1008 merged 18 minutes later; it now exists at `test/e2e/dbt-mixing-e2e.sh:214`. The citation stands.
- **`values-ha.yaml`** does exist — at `helm/leoflow/examples/values-ha.yaml`. Round 2 had the path wrong, round 3 corrected it, and I have used the real path.

Rebased past #1008, so the "if #1008 has not merged" caveat is gone. All §N.M references resolve and every §6 row has a procedure — both checked mechanically.

Issue #1007, opened 2026-09-08
ci: service containers pull from public.ecr.aws, which rate-limits by source IP — ~13% of main runs go red


neochaotic:
## The measurement

Last 30 `ci.yaml` runs on `main`: **4 failures, 26 successes — a ~13% red rate.**

The failures are spread across three different jobs, not concentrated in one:

```
1  Integration tests
1  E2E split (k3d two-process api/scheduler)
1  E2E dbt (k3d pod-per-model)
```

That spread is the tell: it is not a flaky job, it is a shared dependency.

## The cause

The workflows reference `public.ecr.aws` **20 times** — `postgres:16` in 12 places and `redis:7` in 8 — all as **service containers**:

```yaml
services:
  postgres:
    image: public.ecr.aws/docker/library/postgres:16
  redis:
    image: public.ecr.aws/docker/library/redis:7
```

`public.ecr.aws` rate-limits anonymous pulls **by source IP**, and GitHub-hosted runners share IP ranges with everyone else on the platform. So the failure probability has nothing to do with our code, our schedule, or the job — it tracks how much the rest of GitHub is pulling from ECR at that minute.

A representative failure (`run 34290056907`, `E2E split`):

```
/usr/bin/docker pull public.ecr.aws/docker/library/redis:7
Error response from daemon: toomanyrequests: Rate exceeded
Warning: Docker pull failed with exit code 1, back off 2.307 seconds before retry.
... (two more retries)
Error: Docker pull failed with exit code 1
```

Note the retry already exists and is not enough — and it is Docker's own retry, because a **service container is pulled by the runner before any step of ours runs**. We have no hook to wrap it.

## Why it costs more than 13%

With ~10 jobs per run using a service container, the chance that *some* job in a run goes red is well above the per-job rate. And each one costs a rerun plus the "is this a flake or a real failure?" question — which is exactly the ambiguity that makes a red CI easy to wave through.

`FLAKE_RE` in `scripts/cut-release.sh` already classifies `toomanyrequests|Rate exceeded` as transient, so a release cut reruns on its own. That is mitigation of the symptom in one place; PRs get the noise raw.

## Options

1. **Mirror `postgres:16` and `redis:7` into our own `ghcr.io` namespace and point the service containers there.** ghcr does not rate-limit public pulls the way ECR does, and we already publish there. Removes an external dependency from the critical path of every CI run. **Recommended** — it is the only option that removes the cause rather than softening the symptom.
[...]

Issue #978, opened 2026-09-07
cut-release: an unavailable --log-failed reads as "not a flake", so the cut stops on a known transient


neochaotic:
## Observed on the v0.4.5-rc.1 cut

The release-watch loop reported:

```
warn: release workflows red (non-flake) — inspect the run + gate
```

The failure was `/readyz never responded after 60s`, and **`readyz never responded` is in `FLAKE_RE`** — it is there precisely for this, documented as "the lite cold-start /readyz timing flake". Checked after the fact:

```
$ gh run view 34151534260 --log-failed | grep -cE 'readyz never responded'
2
$ gh run view 34151534260 --log-failed | grep -qE "$FLAKE_RE" && echo matches
matches
```

So the pattern was right and the log does match. The classifier still said non-flake.

## The defect

`scripts/cut-release.sh`, both classification sites:

```bash
isflake=1
for rid in $failed; do
  gh run view "$rid" --log-failed 2>/dev/null | grep -qE "$FLAKE_RE" || isflake=0
done
```

`2>/dev/null` plus a pipe means **a failed or empty `gh run view` is indistinguishable from a log that does not match.** GitHub's log endpoint is not reliably available the instant a run completes — observed twice in one session: `gh run view --log-failed --job <id>` returned nothing for jobs that had genuinely failed, while `gh api …/actions/jobs/<id>/logs` returned the full log.

So the loop defaults to "this is a real failure" whenever it cannot read the log, which is exactly the moment it is asked.

This is the same class as #974 and the `check-values-doc-comments` lesson: **a swallowed command failure read as a substantive verdict.**

## Fix

Separate "cannot read" from "read and did not match", and retry the read rather than concluding from it:

```bash
classify_run() { # <run-id> -> flake | real | unknown
  local log
  log=$(gh run view "$1" --log-failed 2>/dev/null) || { echo unknown; return; }
  [ -n "$log" ] || { echo unknown; return; }
  printf '%s' "$log" | grep -qE "$FLAKE_RE" && echo flake || echo real
}
```

`unknown` re-polls (bounded by the existing deadline) instead of deciding. Falling back to `gh api repos/{owner}/{repo}/actions/runs/{id}/logs` on an empty read is worth considering, since that endpoint answered when the other did not.

The pure part — given log text, is it a flake — is table-testable next to the `FLAKE_RE` cases already in `self_test()`; `check-script-selftests.sh` picks it up with no wiring.

## Why this is pre-GA

[...]

Issue #862, opened 2026-09-01
release workflow: retract-to-draft deadlocks download-based smokes on rerun (v0.4.3 got stuck as a draft)


neochaotic:
**Found cutting v0.4.3 GA.** A transient flake retracted the release to a draft, and the rerun could never un-stick it — a workflow deadlock, not a product problem.

## What happened
1. GoReleaser published v0.4.3 (non-draft) and all core artifacts (binaries, server image, migrate image, signature) succeeded.
2. Two transient failures: `runtime-image (py3.12)` hit a `go mod download` / `proxy.golang.org` `INTERNAL_ERROR`; `lite-cold-start-smoke` hit a `/readyz`-in-60s timeout (timing flake — the identical binary passed this smoke at v0.4.3-rc.1).
3. The gate (`.github/workflows/release.yaml:436-443`) retracted the release to a **draft** (`gh release edit --draft`).
4. On `gh run rerun --failed`: `runtime-image` passed, but `lite-cold-start-smoke` now failed with `curl (22) 404 downloading leoflow_0.4.3_linux_amd64.tar.gz` — because the smoke installs via `curl -fsSL .../install.sh | sh` which resolves the tarball from the **release**, and a **draft release's asset URLs 404**. So the smoke can never pass while drafted → the gate's un-draft-on-pass path (release.yaml:446-457, #839) is **unreachable** for any download-from-release smoke. Deadlock.

Manual recovery required: `gh release edit v0.4.3 --draft=false` then rerun.

## Two distinct bugs
1. **Deadlock:** retract-to-draft breaks the rerun of any smoke that downloads from the release (install-smoke, lite-cold-start-smoke, upgrade-smoke all use `curl|sh` / install.sh). Options: (a) don't draft — gate on the jobs and only *promote/latest* on full success, leaving assets downloadable during reruns; (b) have the gate un-draft *before* the smoke jobs on a rerun; (c) point the smokes at the run's artifacts instead of the published release.
2. **Cold-start smoke is timing-flaky:** `lite-cold-start-smoke` polls `/readyz` for 60s, but first boot provisions a dev venv (`pip install` Airflow SDK) which is env/pip-speed dependent and can exceed 60s on a slow runner. Confirmed locally the 0.4.3 binary boots+migrates+starts server/agent correctly; the 60s budget is the fragile part. Options: pre-warm the venv in `leoflow setup` (already a step), raise/relax the budget, or make `/readyz` independent of venv provisioning.

## Impact
[...]

PR #1017, opened 2026-09-09
fix(release): a transient flake during a cut no longer costs a tag

### LINKED ISSUE #979: release gate: retracting to draft breaks the smoke job whose green re-run would un-draft it
## A deadlock, observed live on v0.4.5-rc.1

The `gate` job's success step is explicit about its purpose:

```
# On a green RE-RUN after an earlier run retracted the release to a draft
# (e.g. a transient cosign download flake failed a smoke job), this un-drafts
# it — closing the gap where a passing re-run left the release stuck as a draft.
```

That path is unreachable for every smoke job that installs from the release, which is most of them.

**Sequence, all verified:**

1. `lite-cold-start-smoke` failed on a slow Postgres image pull (#977).
2. `gate` retracted `v0.4.5-rc.1` to a draft. Correct behaviour.
3. Re-ran the failed jobs. `lite-cold-start-smoke` installs with `curl -fsSL …/install.sh | sh`, and `install.sh` resolves the **newest public release** and downloads its asset. A draft release's assets are not public:

```
$ curl -s -o /dev/null -w '%{http_code}' -L \
    https://github.com/neochaotic/leoflow/releases/download/v0.4.5-rc.1/leoflow_0.4.5-rc.1_linux_amd64.tar.gz
404
```

4. Attempt 2 therefore failed with a *different* error — `curl: (22) The requested URL returned error: 404` — and the gate retracted again.

So the retraction removes the input the recovery needs. The only way out is to un-draft by hand, which is precisely the state the gate exists to prevent, or to cut a new rc.

Affects `install-smoke` (7 distro variants), `lite-cold-start-smoke` and `upgrade-smoke` — everything downstream of `install.sh`.

## And the cut would rerun it forever

`returned error: 404` is in `FLAKE_RE` (one of the six grandfathered patterns in #956, none of which has a captured sample). So this **permanent** 404 is classified transient, and `wait_sha_green`'s release-watch loop would rerun it up to 8 times, ~12 minutes each, before reporting RED. That is the concrete harm #956 was speculating about.

## Fix

The smoke jobs should not resolve the release through its public listing. Options, cheapest first:

1. **Install from the built artifact directly.** The `release` job already has the tarball; upload it as a workflow artifact and have the smoke jobs `download-artifact` + install locally. This also makes the smoke test what CI built rather than what the API happens to serve, closing a real gap: today the smoke could pass against a *previous* release if resolution went wrong.
2. **`gh release download --repo … <tag>` with `GH_TOKEN`**, which can read a draft. Smaller change, keeps the install path, but the job still tests the API's answer rather than the artifact.
3. **Retract only after all smoke jobs report**, rather than on the first failure. Reduces the window but does not remove it, since the gate runs once and a re-run still starts from a drafted release.

(1) is the right one: it removes the coupling entirely, and `install.sh` can still be exercised end-to-end against a local file with `LEOFLOW_ARCHIVE=` or equivalent.

Also fix `returned error: 404` as part of #956 — a 404 from our own release URL is not a transient.

## Why this is pre-GA

On a GA, `install.sh` resolves the newest public release, so a retracted GA means `curl | sh` keeps serving the previous version while the tag exists — and per the above, no re-run can recover it. The release would be stuck as a draft until someone un-drafts by hand, which is exactly the manual step the gate was written to remove.

### LINKED ISSUE #983: release: install-smoke and cosign verify have no retries, and any one of them retracts the release
Two sets of steps in `.github/workflows/release.yaml` do unretried network work inside jobs the release gate depends on. Any single flake retracts the pre-release — and because of #979, a retraction cannot be recovered by a re-run.

## 1 · `install-smoke` × 7 distros

`release.yaml:248-278` installs prerequisites with each distro's package manager. `apt-get` runs with `Acquire::Retries=3`; `dnf`, `zypper`, `pacman` and `apk` have no retry configured.

`pacman -Sy --noconfirm curl tar` on `archlinux:latest` is additionally an unsupported partial upgrade (`-Sy` without `-u`), which breaks on mirror rotation rather than on our changes.

Seven jobs, each an independent chance to retract. All seven passed on rc.1, so this is a tail risk rather than an expectation — but the cost of the tail is a new tag.

## 2 · `verify-server-signature`

`release.yaml:405-433` runs `cosign verify` with no retry. It fetches the Sigstore TUF root and talks to Rekor and Fulcio. The release gate's own success-path comment already names this failure:

> *"On a green RE-RUN after an earlier run retracted the release to a draft (e.g. a transient cosign download flake failed a smoke job) …"*

So the scenario is documented and unmitigated, and per #979 the re-run that comment describes cannot succeed.

## Fix

A small retry wrapper on each network step — three attempts, ten seconds apart, as #980 added to the Postgres pre-pull. For `pacman`, either `-Syu` or pin to a dated Arch snapshot.

Worth pairing with #979: retries reduce how often the deadlock is reached, and #979 removes the cost when it is.

neochaotic:
Closes #979, closes #983.

## The deadlock

The gate retracts a failing release to a draft. Most smoke jobs install with `curl … install.sh | sh`, and `install.sh` downloads from the **public** release-asset URL — which a draft does not serve. So the re-run the gate's own success step was written to handle:

> On a green RE-RUN after an earlier run retracted the release to a draft (e.g. a transient cosign download flake failed a smoke job), this un-drafts it

…can never happen. Every re-run fails on the asset download instead, with a failure that looks identical to the original one. Tags are immutable (ADR 0033), so the only exit was a new tag. **Observed live on `v0.4.5-rc.1`.**

## Why this is not self-healing, deliberately

The un-draft has to run in a job that actually re-runs. `gh run rerun --failed` re-runs failed jobs and their dependents, so a new job placed *before* the smoke jobs would have succeeded on the first attempt and be skipped on the re-run. A full re-run would work, but it re-enters GoReleaser against an existing release, and `.goreleaser.yaml` sets no `mode: replace` — I did not want to find out how that goes during a cut.

So this makes recovery **cheap and obvious** rather than automatic:

- The gate writes the two-command recovery to the step summary, and says why a plain re-run cannot work — the reader has no way to deduce that from the failure they are looking at.
- A `workflow_dispatch` workflow re-publishes a named tag in one click. It refuses anything that is not a `vN.N.N` tag, and is a no-op on an already-published release.

Full automation is a real option, but it is a design decision about re-run semantics and belongs with you, not inside a patch.

## Then the probability side (#983)

Every unretried network step in a gating job is an independent chance to retract the release:

| step | before | after |
|---|---|---|
| `cosign verify` | no retry — fetches the Sigstore TUF root, talks to Rekor and Fulcio | 3 attempts, 10s apart |
| 7 distro prereq installs | only `apt-get` had `Acquire::Retries`; `dnf`, `zypper`, `pacman`, `apk` had none | the whole prep retries, 3 × 10s |
| `pacman -Sy` | unsupported **partial upgrade** — breaks on Arch mirror rotation, not on anything we changed | `-Syu` |

The retry shape matches #980's Postgres pre-pull.

## Verification

- `actionlint` clean on both workflows.
- The retry loop was run in a real `sh` against three cases: succeeds first try (no retries, no output), fails always (3 attempts then propagates rc=1), and succeeds on the third (2 retries then continues). Worth doing rather than eyeballing — a broken retry loop in this workflow would only surface during an actual cut, which is the worst place to discover it.

## What this does not fix

[#1007](https://github.com/neochaotic/leoflow/issues/1007) — the service-container pull from `public.ecr.aws`, ~13% of main runs and three hits today. That one fails in `Initialize containers`, before any step, so no retry we write can touch it; it needs a mirror. It is the largest remaining source of the flakes this PR is mitigating the consequences of.


neochaotic:
Review found this PR was building a **second** recovery mechanism next to one the repo already had. Reworked; the new workflow is deleted.

`scripts/cut-release.sh` has done this since #862, and `RELEASING.md` documents it:

```bash
gh release edit "$tag" --repo "$REPO" --draft=false >/dev/null 2>&1 || true
```

**It sits inside `if [ "$isflake" = 1 ]`, and the deadlock defends itself against that test.** Once the gate retracts, the smokes stop failing on whatever failed first and start failing on install.sh's asset download — and `curl -fsSL` prints nothing on a 404, so install.sh only emits `downloading <archive> failed`, which matches nothing in `FLAKE_RE`. `isflake` flips to 0, the watch stops at "release workflows red (non-flake)", and the un-draft below it never runs. **The symptom the deadlock creates is exactly what disables the recovery.** It now re-publishes *before* classifying, which costs nothing when the release is already published.

That is the whole #979 fix, and it is one line of ordering. I should have looked for the existing mechanism before writing a new one.

## A second hole in the same three lines, and it is the one that bites daily

`isflake` treated an **unretrievable** log as evidence of a non-flake:

```bash
gh run view "$rid" --log-failed 2>/dev/null | grep -qE "$FLAKE_RE" || isflake=0
```

`FLAKE_RE` already contains `toomanyrequests|Rate exceeded`, so the #1007 flake *would* be caught — if the log were read at all. It is not: a job that dies in `Initialize containers` has no step log, `--log-failed` returns **empty**, the grep fails, and the cut concludes "real failure" and stops. So the classifier is blindest on the transient it most needs to classify. That is #978, and it is the same empty-measurement trap I walked into by hand this morning.

Now: falls back to the job logs API, and treats a still-unreadable log as **unknown** — which for a cut means rerun, because a rerun is cheap and a stopped cut on a transient is not.

Extracted as `flake_verdict()` next to the other pure logic, with four self-tests including an explicit assertion that empty and a genuine failure do not classify alike. Mutation-verified: flipping the empty case back to `real` reddens both assertions. `_eq` prints only on failure, so I confirmed they run rather than trusting the silence.

## The other findings

- **`pacman -Syu` was the wrong fix and probably worse.** The dominant `archlinux:latest` failure is a stale `archlinux-keyring` ("signature is unknown trust"), and a full upgrade is *more* exposed to it than `-Sy` of two packages — deterministic, so my retry would burn three attempts and retract anyway. Now refreshes the keyring first.
- **cosign** goes through `scripts/cosign-retry.sh`, which every `cosign sign` already uses, instead of a second inline policy at a different backoff. The job gains a checkout, which is why the inline loop existed.
- **`curl … | sh` swallows a failed fetch.** Pre-existing, surfaced by the review. Under `shell: sh` the pipeline reports `sh`'s status, and `sh` succeeds on empty input. Confirmed by running it: a 404 through that pipe exits **0**. Both install steps now fetch and run separately.
- **`RELEASING.md`** updated with both the ordering and the unknown-log behavior.

## Recorded so it is not re-litigated

Retracting with `--latest=false` instead of `--draft` does **not** work: `install.sh` falls back to listing non-draft releases and taking `sort -V | tail -1`, so only drafting removes a bad rc from resolution.

PR #1018, opened 2026-09-09
ci: pull service containers from a registry that does not rate-limit by IP

### LINKED ISSUE #1007: ci: service containers pull from public.ecr.aws, which rate-limits by source IP — ~13% of main runs go red

neochaotic:
Closes #1007.

`public.ecr.aws` returns `toomanyrequests: Rate exceeded` on roughly 13% of main runs. The limit is **per source IP**, and GitHub-hosted runners share a small pool across every customer — nothing about this repo's traffic controls it. Three hits today, one on a PR that changed only markdown, one on `main` right after a merge.

## Why it has been absorbed rather than fixed

It is nearly invisible:

- The pull happens in **`Initialize containers`** — before `actions/checkout`, before any step. So it fires regardless of what the PR touches, and no retry, cache or guard inside the job can reach it.
- It fails in ~14 seconds, which reads as "rerun and move on".
- `gh run view --log-failed` returns **empty** for it, with no error, because a job that dies there has no step log. The obvious triage command reports zero matching lines — easy to read as "no rate limit found" ([#978](https://github.com/neochaotic/leoflow/issues/978)). I did exactly that this morning.

## The change

All 20 references → `mirror.gcr.io/library/<image>`, Google's unauthenticated pull-through cache for Docker Hub official images. Verified both manifests resolve there (HTTP 200) before swapping.

**This trades one third party for another and the PR should say so plainly.** What changes is the failure mode: `mirror.gcr.io` does not impose the shared-source-IP anonymous limit that is actually biting. It is not a guarantee, it is a different and empirically much quieter dependency. If you would rather own the mirror, the alternative is pushing these to GHCR — I did not take it because the package would need a one-time manual visibility flip that `GITHUB_TOKEN` cannot perform, which makes the change land half-broken.

## The guard

`scripts/check-service-image-registry.sh`, globbed into both the CI self-test job and `cut-release.sh`'s `run_gates()`, so no wiring rots.

Two things I made sure of, because a guard that cannot fail is worse than none:

- **It fails on the tree as it stood before this commit** — exit 1, pointing at all 20 lines. Measured, not assumed.
- Its self-test asserts a **comment** naming the registry does *not* trip it. This script and the issue notes name `public.ecr.aws` on purpose, and a gate that cannot survive being documented gets deleted.

## Not affected

`release.yaml`'s Lite pre-pull uses plain `postgres:16`, so `check-lite-prepull-matches-compose.sh` still compares what it compared. `MIN_GATES` is a `-ge` floor, so an added gate does not trip it.

Companion to #1017, which mitigates the *consequences* of a flake during a cut. This one removes the largest source.

### DIFF
--- a/.github/ci/kind-datastores.yaml
+++ b/.github/ci/kind-datastores.yaml
@@ -35,7 +35,7 @@ spec:
     spec:
       containers:
         - name: postgres
-          image: postgres:16-alpine
+          image: mirror.gcr.io/library/postgres:16-alpine
           env:
             - name: POSTGRES_USER
               value: leoflow
@@ -87,7 +87,7 @@ spec:
     spec:
       containers:
         - name: redis
-          image: redis:7-alpine
+          image: mirror.gcr.io/library/redis:7-alpine
           ports:
             - containerPort: 6379
           readinessProbe:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -142,7 +142,7 @@ jobs:
     timeout-minutes: 20
     services:
       postgres:
-        image: public.ecr.aws/docker/library/postgres:16
+        image: mirror.gcr.io/library/postgres:16
         env:
           POSTGRES_USER: leoflow
           POSTGRES_PASSWORD: leoflow
@@ -151,7 +151,7 @@ jobs:
         options: >-
           --health-cmd pg_isready --health-interval 5s --health-timeout 3s --health-retries 10
       redis:
-        image: public.ecr.aws/docker/library/redis:7
+        image: mirror.gcr.io/library/redis:7
         ports: [6379:6379]
         options: --health-cmd "redis-cli ping" --health-interval 5s --health-timeout 3s --health-retries 10
 
@@ -203,7 +203,7 @@ jobs:
     timeout-minutes: 20
     services:
       postgres:
-        image: public.ecr.aws/docker/library/postgres:16
+        image: mirror.gcr.io/library/postgres:16
         env:
           POSTGRES_USER: leoflow
           POSTGRES_PASSWORD: leoflow
@@ -212,7 +212,7 @@ jobs:
         options: >-
           --health-cmd pg_isready --health-interval 5s --health-timeout 3s --health-retries 10
       redis:
-        image: public.ecr.aws/docker/library/redis:7
+        image: mirror.gcr.io/library/redis:7
         ports: [6379:6379]
         options: --health-cmd "redis-cli ping" --health-interval 5s --health-timeout 3s --health-retries 10
     env:
@@ -244,7 +244,7 @@ jobs:
     timeout-minutes: 15
     services:
       postgres:
-        image: public.ecr.aws/docker/library/postgres:16
+        image: mirror.gcr.io/library/postgres:16
         env:
           POSTGRES_USER: leoflow
           POSTGRES_PASSWORD: leoflow
@@ -253,7 +253,7 @@ jobs:
         options: >-
           --health-cmd pg_isready --health-interval 5s --health-timeout 3s --health-retries 10
       redis:
-        image: public.ecr.aws/docker/library/redis:7
+        image: mirror.gcr.io/library/redis:7
         ports: [6379:6379]
         options: --health-cmd "redis-cli ping" --health-interval 5s --health-timeout 3s --health-retries 10
     steps:
@@ -328,7 +328,7 @@ jobs:
     timeout-minutes: 15
     services:
       postgres:
-        image: public.ecr.aws/docker/library/postgres:16
+        image: mirror.gcr.io/library/postgres:16
         env:
           POSTGRES_USER: leoflow
           POSTGRES_PASSWORD: leoflow
@@ -363,7 +363,7 @@ jobs:
     timeout-minutes: 15
     services:
       postgres:
-        image: public.ecr.aws/docker/library/postgres:16
+        image: mirror.gcr.io/library/postgres:16
         env:
           POSTGRES_USER: leoflow
           POSTGRES_PASSWORD: leoflow
@@ -398,7 +398,7 @@ jobs:
     timeout-minutes: 15
     services:
       postgres:
-        image: public.ecr.aws/docker/library/postgres:16
+        image: mirror.gcr.io/library/postgres:16
         env:
           POSTGRES_USER: leoflow
           POSTGRES_PASSWORD: leoflow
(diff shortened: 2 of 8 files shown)


neochaotic:
Empirical confirmation, because "CI is green" would also be true if the service-container jobs had been skipped. From the `Integration tests` job on this PR (`102315559544`):

```
##[group]Starting postgres service container
##[command]/usr/bin/docker pull mirror.gcr.io/library/postgres:16
Status: Downloaded newer image for mirror.gcr.io/library/postgres:16
##[command]/usr/bin/docker create --name …_mirrorgcriolibrarypostgres16_… --health-cmd pg_isready …
##[command]/usr/bin/docker pull mirror.gcr.io/library/redis:7
```

Both images pulled, **first attempt, no retries, no `toomanyrequests`**, and the container was created and used by the job. `Initialize containers` — the step that has been failing — completed in about 7 seconds.

45 checks green, including every job that takes a Postgres or Redis service container.

neochaotic:
Review found the guard was close to inert **and** that its self-test asserted a property the code did not have. Both confirmed by measurement, and both fixed.

I built a fixture with five real banned references and ran the old guard:

```
=> 5 referencias evasivas reais: exit 0
=> um comentario realista:       exit 1
```

It waved through an env indirection, a `${{ }}` expression, a folded scalar, a job-level `container:` and an uppercase `IMAGE:` key — and failed on a comment that merely mentioned the registry. The exact inverse of what its header and my commit message both claimed.

**The self-test passed because my comment fixture happened not to contain the literal `image:`.** That is the third test in this batch that passed for a reason unrelated to what it claimed to check, and this one was in a gate whose entire job is catching drift.

## What it does now

Parses the workflow YAML — the shape `check-lite-prepull-matches-compose.sh` already documents after a reviewer reproduced this same fail-open class in one line. It walks `jobs.*.container` and `jobs.*.services.*.image`, resolves `${{ matrix.<key> }}` against literal matrix values, and **refuses to report clean when the scan target is missing or empty** (the old one printed "ok" for a nonexistent directory — that is how a gate silently stops gating after a move).

It is an **allowlist** now. The header claimed one while the code was a denylist, and the difference is not academic: `image: postgres:16` is bare Docker Hub, rate-limited by the same shared-source-IP mechanism (100 pulls / 6h per IPv4 or /64 — tighter than ECR Public), and pasting it out of `docker-compose.dev.yaml` is the likeliest drift. `quay.io` is allowed: different registry, different pool.

## Parsing found what grepping could not, in the worst possible place

```
release.yaml: jobs.install-smoke.container = debian:12        (…and 5 more distros)
release.yaml: jobs.upgrade-smoke.container = ubuntu:24.04
```

Eight containers pulled from anonymous Docker Hub in `Initialize containers`, **in the workflow where a red job retracts the release and costs a tag** — the same failure class this PR is about, on the path where it is most expensive, and my original swap missed every one. Verified each resolves on the mirror (HTTP 200) before swapping. `.github/ci/kind-datastores.yaml` moved too.

## On the mirror choice — recording the trade honestly

The review is right that Google documents `mirror.gcr.io` for **daemon-level configuration**, not direct addressing. That mode is unusable here: the pull happens before any step we control, so there is nowhere to configure the daemon. So the honest framing is that the only workable form of this option is the undocumented one, there is no SLA, and 20+ call sites now concentrate on one host — the failure mode changes from ~13% red to all-or-nothing. Written into the CHANGELOG rather than left implicit.

The durable answer is a mirror we own. I also need to correct something I wrote in #1007: I said there was no place to authenticate before the service-container pull. That is **wrong** — `jobs.<id>.services.<id>.credentials` is a first-class documented field. What actually blocks GHCR-private is fork PRs (read-only repo-scoped token, long-standing reports of private-package pulls being denied), which argues for a **public** GHCR package, whose only real cost is the one-time manual visibility flip GitHub's API cannot perform.

**Merge order:** #1017 also edits the `install-smoke` matrix lines (the `prep` strings). Whichever lands second needs a trivial rebase; I will take it.

neochaotic:
Rebased past #1017. The conflict was in the `install-smoke` matrix, where both PRs edit the same lines — #1017 the `prep` strings, this one the `image` values. Resolved keeping **both**, and verified each half survived rather than assuming:

- `archlinux-keyring` refresh from #1017 — present
- 6 mirrored `image:` values — present
- `flake_verdict` / the un-draft ordering in `cut-release.sh` — present
- `cosign-retry.sh verify` and the split `curl`/`sh` install steps — present

The `CHANGELOG.md` merges were purely additive both times (`git diff origin/main -- CHANGELOG.md | grep -c "^-[^-]"` → 0).

Guard, both self-test suites and `actionlint` clean on the rebased tree.

Issue #860, opened 2026-09-01
helm: control-plane defaults + un-tunable 1s liveness timeout can get the server kubelet-killed under load


neochaotic:
**Field feedback (#8)** from the 2026-09-01 EKS incident. Confirmed in code at `b7dd0f8`.

## What
Chart control-plane defaults are tiny and the probe timeout is not operator-tunable:
- `helm/leoflow/values.yaml:392-398`: requests `100m/128Mi`, limits `1/512Mi`.
- `helm/leoflow/templates/_controlplane-deployment.tpl:333-338`: liveness/readiness `httpGet /healthz` `/readyz`, `initialDelaySeconds:10, periodSeconds:20`; `timeoutSeconds`/`failureThreshold` **omitted** → Kubernetes defaults **1s / 3**. No value exposes them — an operator cannot raise the timeout without editing the template (a fork).

`livenessHandler` is trivially cheap (static JSON, no DB — `internal/api/health.go:16-18`), so a liveness failure is **pure CPU starvation** of the process. Under CFS throttling at `limits.cpu:1` during a task-pod fan-out, the server can't answer `/healthz` within 1s; ~3 failures × 20s ⇒ ~40-60s of sustained throttle → kubelet restart → triggers the agent-lost cascade (#858/#859).

## Fix (next RC)
- Expose `livenessProbe`/`readinessProbe` `timeoutSeconds`/`periodSeconds`/`failureThreshold` in `values.yaml`.
- Raise default liveness `timeoutSeconds` to 3-5, keep `failureThreshold:3`.
- Raise default requests (a control plane fanning out pods should request ≥250m/256Mi).

Note: bumping defaults is the shallow mitigation; #858/#859 are the real robustness fix so a restart (for any reason) no longer fails healthy runs. Operator already mitigated in prod (control plane → 1-2 CPU / 512Mi-1Gi).

PR #1033, opened 2026-09-09
fix(api): make /readyz assert the schema invariant boot asserts, on every probe (#1023)

### LINKED ISSUE #1023: /readyz reports ready with an empty database: the schema check exists at boot but not at readiness
Found during the v0.4.5 RC cluster validation on GKE. This is the kind of thing the runbook exists to find, and it is not in the runbook.

## What happened

The validation cluster's Postgres runs on ephemeral storage. A node recycle recreated the pod and the database came back **empty**. The control plane pod kept running, and:

```
$ curl -sS http://…/readyz
{"status":"ready"}     HTTP 200

$ psql -tAc "select count(*) from information_schema.tables where table_schema='public'"
0
```

Zero tables. Meanwhile, in the same seconds:

```
{"level":"ERROR","msg":"scheduler step","error":"listing active runs: ERROR: relation \"dag_runs\" does not exist (SQLSTATE 42P01)"}
{"level":"ERROR","msg":"http request","path":"/auth/token","status":503,"detail":"authentication temporarily unavailable"}
{"level":"INFO","msg":"http request","path":"/readyz","status":200}
```

The scheduler cannot complete a single step, nobody can authenticate, and the pod reports itself **Ready**.

## Root cause

`readinessHandler` (`internal/api/health.go:20-35`) iterates the dependency checkers and calls `Ping`. A Postgres `Ping` succeeds whenever the *connection* is healthy; it says nothing about the schema.

The invariant is enforced — but only at startup. `cmd/leoflow-server/main.go` refuses to boot without it, with an excellent message:

```
leoflow-server: postgres: database schema is not current: schema_migrations is absent
— migrations have not run …; schema v26 is required
```

So the server knows the rule, checks it once, and then never re-checks. A database that changes underneath a **running** pod is invisible to readiness.

## Why this matters beyond my ephemeral-Postgres setup

`/readyz` is what Kubernetes uses to decide whether to route traffic and whether a rollout succeeded. With this gap:

- **`helm upgrade --wait` reports success** against a control plane that cannot serve a single authenticated request. That is the failure mode I actually hit — the install looked healthy.
- A **rolling update completes**, replacing working pods with broken ones, because each new pod passes readiness.
- The Service keeps the pod **in rotation**, so a load balancer sends real traffic to something that 503s.

Reachable outside a dev cluster: a restore from a backup taken before the current schema, a failover to a replica that has not caught up, a migration rolled back out of band, or a `database.url` repointed at the wrong database — the last one is not hypothetical either, since the URL is a plain chart value.

## Suggestion

Readiness should assert the same invariant boot does: `schema_migrations` present and at the expected version. The check is a single indexed query and readiness already tolerates a per-dependency probe. Keeping the response body vague is right (the existing comment about not leaking a DSN in an unauthenticated endpoint still applies) — the log carries the detail.

Worth deciding deliberately whether **liveness** should also fail. My instinct is no: restarting does not create a schema, and a crash loop is worse than a pod that is honestly not-ready. But readiness must not claim ready.

## Runbook

Whatever the fix, §3 or §4 should gain a row: drop the schema (or point at an empty database) under a running control plane and assert `/readyz` goes non-200. It is a two-command check and it would have caught this.

neochaotic:
Closes #1023.

## The bug

`/readyz` answered `{"status":"ready"}` HTTP 200 against a database with zero tables, while the scheduler was failing every tick on `relation "dag_runs" does not exist` and `/auth/token` was answering 503. `readinessHandler` called `Ping` on each dependency and nothing else, and a Postgres `Ping` succeeds whenever the *connection* is healthy — it says nothing about the schema.

The invariant was already enforced, but only at boot, where the server refuses to start ("`schema_migrations` is absent — migrations have not run; schema v26 is required"). A boot check cannot see a database that changes under a running pod. Since `/readyz` is what Kubernetes routes traffic on and judges a rollout by, this let `helm upgrade --wait` report success against a control plane that could not serve one authenticated request.

## The fix

Readiness asserts the same invariant, through the same code path:

- `storage.(*Postgres).schemaState` is new and unexported, and is now the single place both the boot gate (`CheckSchemaCurrent`) and the probe read the DB version and the binary's embedded latest. **The version constant is not duplicated** — it stays `migrations.Latest()` — and the two verdicts cannot drift.
- `storage.(*Postgres).SchemaReady` is the probe's entry point. It differs from `CheckSchemaCurrent` in exactly one respect: no ahead-schema warning, which is boot-only reporting and would otherwise repeat once per probe period for as long as the condition lasted. An **ahead** schema still passes, as at boot — expand-contract migrations keep older code working, and failing it would CrashLoopBackOff old pods on `helm rollback`.
- `api.SchemaChecker` is an **optional** capability on `HealthChecker`. Redis and other schema-less dependencies keep the Ping-only contract untouched.
- `var _ api.SchemaChecker = (*storage.Postgres)(nil)` in `main.go`. A type assertion that stops matching is silent — without this pin, a rename on either side returns `/readyz` to reporting ready over an empty database with every test still green.

Three constraints, each with a test that fails without it:

- **The body stays vague.** `/readyz` is unauthenticated and the error can carry a DSN or an internal hostname (audit H2), so the response names the dependency (`postgres schema not current`) and the detail goes to the log — the same split the existing `Ping` path uses.
- **Bounded at 2s**, under the chart's 3s `probes.readiness.timeoutSeconds`. The existing `Ping` path is bounded only by the request context; a schema query with no deadline of its own would turn a probe that should report not-ready into one that reports nothing. `Ping` itself is left as it was — I did not widen the change to it, and it is worth a follow-up decision rather than a silent one here.
- **Ping first, short-circuiting.** With the connection down there is nothing to learn from a schema query, and it would only burn the probe's budget.

## Liveness is unchanged, deliberately

I agree with the issue. Restarting creates no schema, so a crash loop would trade an honestly not-ready pod for one that cannot be inspected — and it would do it to every replica at once, on a condition that is often external and self-healing (a migration Job that has not finished, a replica catching up). The reasoning is now a comment on `livenessHandler` rather than tribal knowledge, and the new runbook row asserts `RESTARTS` does not climb.

## Tests

TDD, and verified as such rather than asserted:

**`internal/api/health_schema_test.go`** (7 subtests): 200 on a current schema; 503 when `schema_migrations` is absent; 503 when the version is behind; the body leaks none of `hunter2` / `pg.internal.svc` / `postgres://` / `5432` out of a DSN-bearing error; the schema call receives a deadline; a Ping-only dependency is unaffected; a dependency that fails Ping is never queried.

Run against the unmodified handler **before** implementing: 5 of the 7 fail. The two that pass are the two asserting *unchanged* behavior (the Ping-only dependency, and no schema call after a failed Ping), which is what they are there for.

**Mutation-tested afterwards**, each restored before the next:

| mutation | result |
| --- | --- |
| `if err != nil` → `if false` on the schema verdict | the 3 status assertions redden |
| drop `context.WithTimeout`, pass the request context | the deadline assertion reddens |
| append `err.Error()` to the problem detail | the leak assertion reddens, naming all 4 secrets |

**`internal/storage/schema_ready_integration_test.go`** (`//go:build integration`): the same states against a real Postgres, because "does this SELECT report an absent table as absent, through this pool" is exactly what a fake cannot answer — which is the shape of the bug being fixed. Each case connects with `search_path` pointed at a scratch schema, so an absent or stale `schema_migrations` is produced **without touching** the migrated schema the rest of the suite shares, and the scratch schemas are dropped in `t.Cleanup`. A fourth case asserts boot and readiness return the *identical* error for the same database — the anti-drift property.

**Honest limitation:** there is no Docker or Postgres on this machine, so the integration file is verified locally only by `go vet -tags integration` and `golangci-lint --build-tags integration` (both clean, and it failed to compile before `SchemaReady` existed). Its first real execution is this PR's `integration` job — please treat a green there as the actual evidence, and I will follow up if it is not.

## Also verified

- `go test ./...` green, `golangci-lint run ./...` **0 issues** (cache cleaned first — a stale cross-worktree cache was reporting findings from other checkouts).
- `scripts/check-changelog-entry.sh`, `check-no-stale-deploy-path.sh`, `check-values-doc-comments.sh`, `check-service-image-registry.sh`, `check-script-selftests.sh` all pass.
- CHANGELOG entry under `## [Unreleased]` (user-facing; `skip-changelog` does not apply).
- RC runbook gains **§4.6**: drop the schema under a running control plane, assert `/readyz` goes 503 and the pod leaves the endpoints, assert it does **not** restart, and assert recovery needs no pod churn.

## Not changed, on purpose

ADR 0010 still says readiness "returns 200 only if dependencies are reachable". It is a record of a decision made then, and this PR supersedes the behavior rather than the history — say the word if the house rule is to amend it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,13 +101,40 @@ adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   a replica that has not caught up, a migration rolled back out of band, and a
   `database.url` repointed at the wrong database.
 
-  Three properties of the fix are deliberate. The response body stays vague —
-  it names the dependency (`postgres schema not current`) and nothing else,
-  because `/readyz` is unauthenticated and the underlying error can carry a DSN
-  or an internal hostname; the detail goes to the log. **Liveness is
-  unchanged**: restarting a pod does not create a schema, so a crash loop would
-  trade an honestly not-ready pod for one that cannot even be inspected. And
-  the check is bounded at 2s, under the chart's 3s
+  **A migration in flight does not take the control plane out of rotation.**
+  The migrate Job is a `pre-install,pre-upgrade` Helm hook, so on upgrade it runs
+  while the old pods are still live and serving, and golang-migrate marks the
+  schema dirty for the whole execution of each migration body. Every replica
+  reads that same row, so a probe that treated dirty as not-ready would flip all
+  of them at the same instant for any migration longer than 30s
+  (`periodSeconds × failureThreshold`) and drop the Service to zero endpoints —
+  and in the single-Deployment `all` role that Service also carries gRPC, so
+  running task pods would lose the control plane mid-upgrade. Readiness is
+  therefore version-aware where boot is not: a dirty schema **above** the
+  version the running binary embeds is a forward migration that does not concern
+  it, and the pod keeps serving (one log line per episode, not one per probe). A
+  dirty schema at or below that version is a half-applied schema the binary
+  actually depends on, and it goes not-ready. Boot still refuses to *start*
+  against any dirty schema — a process that has not begun serving has nothing to
+  lose by waiting, and that is the one state where the two verdicts are meant to
+  differ.
+
+  **`/api/v2/monitor/health` asserts the same thing.** It reads the same
+  dependency map and was also Ping-only, so in the state above it reported
+  `metadatabase: healthy` at HTTP 200 while `/readyz` was 503 and the pod was out
+  of the Service — the Airflow-compatible surface the UI dashboard renders was
+  the last place still claiming the database was fine.
+
+  Three further properties of the fix are deliberate. The response body stays
+  vague — it names the dependency and nothing else, because `/readyz` is
+  unauthenticated and the underlying error can carry a DSN or an internal
+  hostname; the detail goes to the log. It also distinguishes what it saw:
+  `postgres schema not current` is a schema verdict, `postgres unavailable` is a
+  database that could not be read at all (a timeout, a reset connection, an
+  exhausted pool). Both are 503, but only one of them is a reason to go looking
+  at migrations. **Liveness is unchanged**: restarting a pod does not create a
+  schema, so a crash loop would trade an honestly not-ready pod for one that
+  cannot even be inspected. And the check is bounded at 2s, under the chart's 3s
   `probes.readiness.timeoutSeconds`, so a wedged database makes the probe report
   not-ready rather than report nothing at all. An **ahead** schema still passes,
   as it does at boot — expand-contract migrations keep older code working, and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,38 @@ adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   **What to do:** set `python_version: "3.11"` (or `"3.12"` / `"3.13"`) in
   `leoflow.yaml` and rebuild — or, if you pinned `base_image`, repoint it to the
   matching `py3.11` tag and rebuild.
+### Fixed
+
+- **`/readyz` no longer reports ready over a database with no schema (#1023).**
+  The readiness probe pinged each dependency and nothing more, and a Postgres
+  `Ping` succeeds whenever the *connection* is healthy — it says nothing about
+  what is behind it. Found on the v0.4.5 RC cluster: a node recycle recreated an
+  ephemeral-storage Postgres **empty**, and the running control plane answered
+  `{"status":"ready"}` HTTP 200 in the same seconds the scheduler was failing
+  every tick on `relation "dag_runs" does not exist` and `/auth/token` was
+  answering 503. Readiness now asserts the same invariant boot does —
+  `schema_migrations` present, clean, and not behind the version this binary
+  embeds — through the same storage-layer code path, so the two verdicts cannot
+  drift. This is the signal Kubernetes routes traffic on and calls a rollout
+  successful on, so the gap let `helm upgrade --wait` report success against a
+  control plane that could not serve one authenticated request, let a rolling
+  update replace working pods with broken ones, and kept a 503-ing pod in
+  Service rotation. It also covers the cases that reach installs with durable
+  storage: a restore from a backup older than the running binary, a failover to
+  a replica that has not caught up, a migration rolled back out of band, and a
+  `database.url` repointed at the wrong database.
+
+  Three properties of the fix are deliberate. The response body stays vague —
+  it names the dependency (`postgres schema not current`) and nothing else,
+  because `/readyz` is unauthenticated and the underlying error can carry a DSN
+  or an internal hostname; the detail goes to the log. **Liveness is
+  unchanged**: restarting a pod does not create a schema, so a crash loop would
+  trade an honestly not-ready pod for one that cannot even be inspected. And
+  the check is bounded at 2s, under the chart's 3s
+  `probes.readiness.timeoutSeconds`, so a wedged database makes the probe report
+  not-ready rather than report nothing at all. An **ahead** schema still passes,
+  as it does at boot — expand-contract migrations keep older code working, and
+  failing it would break `helm rollback`.
 
 ## [0.4.5] - 2026-09-09
 
--- a/cmd/leoflow-server/health_checks_test.go
+++ b/cmd/leoflow-server/health_checks_test.go
@@ -1,0 +1,76 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/neochaotic/leoflow/internal/api"
+	"github.com/neochaotic/leoflow/internal/storage"
+)
+
+// pingOnlyChecker stands in for the decorator this test exists to catch — a
+// metrics wrapper, a circuit breaker, a tenant router. It satisfies
+// api.HealthChecker, so healthChecks would still compile with it in the map.
+type pingOnlyChecker struct{ api.HealthChecker }
+
+// TestHealthChecksPostgresCarriesTheSchemaAssertion pins #1023 where it can
+// actually break, which is not where a compile-time interface pin can see it.
+//
+// The readiness handler discovers the schema assertion with a runtime type
+// assertion on the VALUE in the checks map. `var _ api.SchemaChecker =
+// (*storage.Postgres)(nil)` pins the TYPE, so it catches a rename on either
+// side and nothing else: wrap the value in any decorator that implements only
+// Ping and the pin still compiles, every test stays green, and /readyz silently
+// returns to reporting ready over an empty database — the exact regression
+// #1023 was. The value is what the handler sees, so the value is what to assert.
+func TestHealthChecksPostgresCarriesTheSchemaAssertion(t *testing.T) {
+	// A nil Pool is fine: nothing here reaches the database, and constructing
+	// the value is the whole point — it is the map entry that must keep the
+	// capability, not the declared type of the field it came from.
+	pg := &storage.Postgres{}
+
+	checks := healthChecks(pg, nil)
+
+	entry, ok := checks["postgres"]
+	if !ok {
+		t.Fatalf(`no "postgres" entry in the readiness checks: %v`, keys(checks))
+	}
+	if _, ok := entry.(api.SchemaChecker); !ok {
+		t.Fatalf("the postgres readiness check (%T) does not implement api.SchemaChecker, "+
+			"so /readyz and /api/v2/monitor/health silently fall back to Ping-only and report "+
+			"ready over a database with no schema (#1023)", entry)
+	}
+}
+
+// TestHealthChecksDecoratorLosingSchemaIsCaught is the negative control: it
+// proves the assertion above can fail, by building the map shape a Ping-only
+// decorator would produce.
+func TestHealthChecksDecoratorLosingSchemaIsCaught(t *testing.T) {
+	wrapped := api.HealthChecker(pingOnlyChecker{&storage.Postgres{}})
+	if _, ok := wrapped.(api.SchemaChecker); ok {
+		t.Fatal("pingOnlyChecker must NOT satisfy api.SchemaChecker, or the guard above proves nothing")
+	}
+}
+
+// TestHealthChecksOmitsRedisWhenAbsent keeps the map's other contract honest:
+// Redis is schema-less and is registered only when it is the active datastore.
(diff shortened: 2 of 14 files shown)


neochaotic:
Review found this PR introduces a **new outage on every `helm upgrade`**. Not merging; sending back with the findings.

## B1 — the dirty window empties the Service (blocking)

`SchemaReady` calls `checkSchemaCurrent` verbatim, and that has `case dirty: return errSchemaNotCurrent`. Meanwhile `migration-job.yaml:18` is `helm.sh/hook: pre-install,pre-upgrade` — on upgrade it runs **while the old pods are live and in the Service endpoints** — and golang-migrate sets `dirty=true`, committed in its own transaction, for the whole execution of each migration body.

So any upgrade whose migration takes longer than `failureThreshold × periodSeconds` = 3 × 10s = 30s flips **every** old replica NotReady at the same instant (they all read the same row), and the Service drops to zero endpoints. In the `all` role that Service also carries gRPC, so running task pods lose the control plane mid-migration — the `agent_lost` cascade #860 exists to prevent.

Second arm: a **failed** migration leaves `dirty=true` permanently. Today that degrades the control plane; after this PR it removes it from rotation entirely, and since boot also hard-fails on dirty, any pod that restarts for an unrelated reason CrashLoopBackOffs. The PR's "no crash loop" promise does not hold in the dirty case.

**The framing is what caused it.** "Boot and readiness cannot drift" is the wrong invariant for `dirty` specifically. Boot-time dirty means *refuse to start against a half-applied schema*. Run-time dirty means *a migration is in flight right now*, and a pod serving correctly a second ago should keep serving. This is the one state where the two verdicts **should** differ — which means sharing `checkSchemaCurrent` verbatim is the bug, not an economy.

Fix shape: `SchemaReady` becomes version-aware on dirty — `dirty && version > latest` is **ready** (an in-flight forward migration past what this binary needs, i.e. exactly the old-pods-during-upgrade case), `dirty && version <= latest` is not ready. Failing test first: an old binary (`latest=26`) against `(version=27, dirty=true)` returns nil from `SchemaReady` and non-nil from `CheckSchemaCurrent`.

## B2 — every transient DB error reads as "schema not current" (blocking, cheap)

`health.go` treats any non-nil return as a schema verdict, but `schemaState` also returns `reading database schema version: <deadline exceeded | conn reset | pool exhausted>`. So the 2s timeout produces **`postgres schema not current`** for a database that is merely slow, sending on-call to migration-land for a pool problem. And §4.6 of the runbook makes that exact string the PASS criterion, so the row passes for the wrong reason.

`errSchemaNotCurrent` already exists as a sentinel — export an `IsSchemaNotCurrent` and branch the detail. Both still 503, both still leak nothing.

## B3 — the pin does not pin what breaks (blocking)

`var _ api.SchemaChecker = (*storage.Postgres)(nil)` stops a rename, correctly. It does **not** stop the failure it is advertised against: the runtime assertion is on the **value in the map**, the pin is on the **type**. Wrap `*storage.Postgres` in any decorator implementing only `Ping` — metrics, circuit breaker, tenant router — and the pin still compiles, tests stay green, `/readyz` silently returns to Ping-only. There is no test on `healthChecks()` at all.

Replace with a test asserting the invariant on the value: `healthChecks(pg, nil)["postgres"].(api.SchemaChecker)`. Failing-test-able today with a nil-`Pool` `*storage.Postgres`, no DB needed.

## B4 — the adjacent call site with the identical bug (should ship here)

`internal/api/monitor.go:47-55` — `/api/v2/monitor/health` reads the **same checks map** and still does Ping-only, reporting `metadatabase: healthy` at HTTP 200. In the exact #1023 state, after this PR: `/readyz` 503, pod out of endpoints, scheduler failing every tick, and the Airflow-compatible surface the UI renders says the metadatabase is fine. One-line application of the same interface.

## B6 — runbook §4.6 has two rows that cannot fail

- `RESTARTS` checked instantly cannot fail: liveness is 20s × 3, so 60s of failures would be needed; the row shows `RESTARTS 0` whether liveness is schema-blind or not.
- `READY 0/1` checked instantly cannot fail the other way: readiness needs up to 30s, so a **correctly fixed** build shows `READY 1/1`. Needs `kubectl wait --for=condition=Ready=false --timeout=60s`.
- The `curl` breaks itself: port-forward through the **Service** drops when the pod leaves the endpoints, so it prints `000`, not `503`. Forward to the pod.
- No §6 results row was added, and that file states a blank row reads as "not run".

## Not blocking, worth issues

The 2s is a per-step deadline, not a probe budget (the handler total is unbounded, though it still fails safe); nothing gates 2s against the chart's operator-settable `timeoutSeconds`; and the probe now takes a second pool acquire against `maxOpenConns: 20`, so under saturation every replica fails readiness simultaneously — a load-induced outage rather than a database one. A dedicated single-conn health pool is the shape `NewLeaderPool` already establishes.

**Confirmed correct and not to be re-litigated:** the #1023 defect is closed; ahead ⇒ ready is right, and for readiness the argument is *stronger* than at boot (failing it would leave `helm rollback` unable to converge); liveness unchanged is right.

neochaotic:
All five findings addressed on the branch (`b5b2c3a..2915c01`). Taking them in the order the review raised them.

## B1 — the dirty window

Fixed as prescribed, and the framing critique is accepted: sharing `checkSchemaCurrent` was the bug, not the economy. Readiness now runs its own gate, `checkSchemaReady`, differing from boot in exactly one state.

| state | boot | readiness |
| --- | --- | --- |
| `!exists` | fail | not ready |
| `version < latest` | fail | not ready |
| `dirty && version > latest` | fail | **ready** |
| `dirty && version <= latest` | fail | not ready |
| clean, `version >= latest` | ok | ready |

The discriminator is one predicate, `dirtyAhead`, used by both the gate and the log line so they cannot come to disagree about which state the database is in. The log is latched per **episode**, not per process: it clears when the condition clears, so a later upgrade is reported again instead of being silenced for the life of the pod.

The divergence is asserted as the point rather than a side effect — `latest=26` against `(27, dirty=true)` returns `nil` from `SchemaReady` and non-nil from `CheckSchemaCurrent`, both at the pure-gate level and, seeded at v9999, against a real Postgres.

## B2 — transient vs schema

`errSchemaNotCurrent` moved to `internal/domain` as `ErrSchemaNotCurrent`, next to the other cross-cutting sentinels, since two layers now need to agree on it. I used `errors.Is` against the exported sentinel rather than adding an `IsSchemaNotCurrent` wrapper — same branch, one less indirection; say the word if you want the named predicate. `context.DeadlineExceeded` now produces `postgres unavailable`; a real verdict still produces `postgres schema not current`. Both 503, both leak nothing, and there is a test asserting neither classification leaks a DSN.

## B3 — the pin

You were right, and I verified it rather than taking it on faith: wrapping the map value in a Ping-only decorator **still builds clean** with `var _ api.SchemaChecker = (*storage.Postgres)(nil)` in place. The pin is replaced by `TestHealthChecksPostgresCarriesTheSchemaAssertion`, which asserts on `healthChecks(pg, nil)["postgres"]` — the value, from the constructor the server calls — plus a negative control proving the guard can fail. There was no test on `healthChecks` at all before.

## B4 — the monitor endpoint

Shipped. Ping and the schema assertion are now one helper shared by `/readyz` and `/api/v2/monitor/health`, so two handlers reading one map cannot disagree. The schema call is bounded on this path too, since the UI polls it.

## B6 — the runbook

All four, plus the recovery step. §4.6a: `kubectl wait --for=condition=Ready=false --timeout=60s`; a ~90s hold with a positive `/healthz` 200 assertion across the window; port-forward to the **pod**, with the reason stated (a Service forward dies with the endpoint and prints `000`, so the row would fail a correct build). Recovery now says the Job is gone under `hook-delete-policy: hook-succeeded` and that `helm upgrade` is what re-fires the hook, and spells out the `ALTER SCHEMA … OWNER TO` / `GRANT` sequence, since `CREATE SCHEMA public` restores the schema but not who may use it.

New **§4.6b** covers the property this review is actually about and that §4.6a cannot observe — a rolling upgrade never dropping the Service to zero endpoints — with a fail-closed second arm. Both have §6 rows.

## Filed, not fixed here

#1040 (per-step deadline vs whole-probe budget), #1041 (`values.schema.json` floor on `probes.readiness.timeoutSeconds`), #1042 (dedicated single-conn health pool). ADR 0010 gets a supersession note rather than an edit to the decision.

## On the evidence, since last time it was thin

You flagged that the integration test does `CREATE SCHEMA`/`DROP SCHEMA` as whoever owns `DATABASE_URL`. Checked: CI's DSN is the `leoflow` role the postgres image bootstraps — `rolsuper = t`, `has_database_privilege(…,'CREATE') = t` — so green there is real. The helper now skips with the reason on `42501` instead of failing, and I confirmed that empirically by creating a read-only role and re-running: the fixture-building cases SKIP, the one that needs no fixture still runs.

And this time the integration suite was **actually executed**, not just vetted — Docker was available, so it ran against a real Postgres 16 with migrations applied at v26. 9 subtests across two tests, all passing, including a single-pool walk through `behind → dirty-ahead → clean → dirty-ahead → table dropped` that pins both that the verdict follows the database and that the log is once per episode.

## Mutations

Each applied alone and reverted before the next.

| mutation | caught by |
| --- | --- |
| `dirtyAhead` → `false` (the blocking behavior) | unit divergence test **and** integration `dirty-ahead` case |
| `dirtyAhead` → `exists && dirty` (widen to "dirty is fine") | unit `dirty at latest`; integration case seeded at `migrations.Latest()` |
| latch never resets | integration: `logged 1 times across two separate episodes, want 2` |
| log per probe instead of latching | integration: `logged 6 times across 6 probes, want 1` |
| `unreadyDetail` always "schema not current" | the `DeadlineExceeded` test |
| `unreadyDetail` always "unavailable" | the real-verdict test |
| monitor reverts to `hc.Ping` | `TestMonitorHealthAssertsSchemaInvariant` (2 subtests) |
| drop `context.WithTimeout` | deadline assertions on both endpoints |
| wrap the map value in a Ping-only decorator | **builds clean**; caught only by the value-level guard |

The first widen-mutation initially survived integration, which is why the `migrations.Latest()`-seeded case and the single-pool walk exist — they were added to close gaps the mutations exposed.

`go build ./...`, `go vet` (both tag sets), `go test ./...`, `-race`, and `golangci-lint run ./...` → **0 issues**. All `scripts/check-*.sh` pass except the release-only one that needs a tag argument. Filtered coverage 81.0% against the 80 floor; the new pure gates are at 100%.

One correction to my earlier PR text: `golangci-lint --build-tags integration ./...` is **not** clean on this repo — it reports 60 pre-existing issues. I verified they are all present at `b5b2c3a` and my files contribute zero. The CI gate runs without that tag, where it is 0.

Issue #1031, opened 2026-09-09
py3.10 base freezes at its 2026-10-31 EOL, and our two supported-Python lists already disagree


neochaotic:
Two problems that resolve together, both measured.

## 1. The py3.10 runtime base stops being rebuilt in seven weeks

Python 3.10 reaches EOL **2026-10-31**. What happens then is not speculation — the `docker-library/python` project stops rebuilding an EOL line the day after:

> `python:3.9-slim-trixie` was last rebuilt **2025-11-01**. Python 3.9 went EOL 2025-10-31.

So from November, `ghcr.io/neochaotic/leoflow-runtime:py3.10` (`.github/workflows/release.yaml:167`) inherits a base that upstream no longer patches, and it accumulates unfixed OS CVEs indefinitely. This is a **larger exposure than the bookworm→trixie suite question**, because a suite bump at least lands on something still maintained.

It is also worse for us than for most projects: `base_image` in `leoflow.yaml` defaults to our published runtime, so a user's DAG image inherits our choice and keeps inheriting it until they rebuild.

## 2. Our two supported-Python lists already disagree, in both directions

| source | versions |
|---|---|
| `.github/workflows/release.yaml:167` (what we publish) | 3.10, 3.11, 3.12 |
| `internal/setup/detect.go:76-77` `pythonCandidates` (what the host detector probes) | 3.11, 3.12, **3.13** |

So the CLI already probes for a 3.13 interpreter we publish no base for, and publishes a 3.10 base the detector will not select. Neither list is wrong on its own; there are just two of them, and nothing reconciles them. `runtime/Dockerfile:13` carries a third statement of the same fact in a comment.

## What the ecosystem actually supports today

Measured from PyPI, current releases:

| package | max Python |
|---|---|
| `apache-airflow-task-sdk` 1.3.1 | 3.14 |
| `dbt-core` 1.12.4 | 3.14 |
| `dbt-postgres` 1.11.0 | 3.14 |
| `dbt-snowflake` 1.12.0 | **3.13** |
| `dbt-bigquery` 1.12.0 | **3.13** |
| `dbt-databricks` 1.12.5 | **3.13** |
| `dbt-duckdb` 1.11.0 | **3.13** |

[...]

Issue #437, opened 2026-07-19
ci(security): run Trivy fs scan on PRs, not just push/schedule


neochaotic:
## Problem

The Trivy filesystem scan is gated to `push` and `schedule` only:

```yaml
# .github/workflows/security.yaml:70
trivy:
  name: Trivy (image scan)
  if: github.event_name == 'push' || github.event_name == 'schedule'
```

So on every PR it reports **`skipping`**, and dependency CVEs Trivy catches are only surfaced **after** merge, on the post-merge `main` Security run. `govulncheck` (which *does* run on PRs) doesn't cover the gap: it reports only vulnerabilities reachable by our call paths, whereas Trivy flags the dependency version regardless of reachability.

## Concrete incident (2026-07-19)

After the #424 alerting epic merged (#430 + #433), the main Security run went **red**. Two fixable HIGH advisories in indirect deps, disclosed *after* the PRs' CI ran, first surfaced on main:

- `golang.org/x/net`  v0.55.0 → CVE-2026-46600 (fixed v0.56.0)
- `golang.org/x/text` v0.37.0 → CVE-2026-56852 (fixed v0.39.0)

Fixed in #436, but only after main was already red — exactly the "PR-green ≠ main-green" trap.

## Proposal

Drop the `if:` so Trivy fs runs on `pull_request` too. It's cheap (~30s scan + DB pull; the fs scan is not the image build), and it moves the signal left so a fixable CVE blocks the PR instead of reddening main.

### Considerations
- **New-CVE-after-PR-CI is still possible.** A CVE disclosed between a green PR and its merge will still land on main. Running on PRs shrinks the window but doesn't eliminate it — the weekly `schedule` run + prompt `main` watching remain the backstop.
- **Noise / flakes.** Trivy DB reclassifications (`WARN: using severities from other vendors`) can flip a finding in/out of the HIGH,CRITICAL filter. On PRs this could intermittently block; keep `ignore-unfixed: true` (already set) so only actionable, fixable findings gate.
- **SARIF upload category.** The upload step is `if: always()` with `category: trivy-fs`; confirm PR-context SARIF uploads don't clobber the push-context baseline in the Security tab (may want a distinct category or skip upload on PRs, scan-only).

## Acceptance
- [ ] Trivy fs runs on `pull_request` to main
- [ ] A PR that introduces/keeps a fixable HIGH/CRITICAL dep CVE fails its own check
- [ ] No SARIF-category collision between PR and push runs in the Security tab

[...]

PR #1034, opened 2026-09-09
ci: scan container images for CVEs, with a policy that can stay on

neochaotic:
`govulncheck` reads our Go module graph. It says nothing about the Debian packages inside `python:3.x-slim-bookworm`, nor about a third-party Go binary baked into someone else's image. That is the gap that let a class of vulnerability reach a human before it reached CI.

This PR is **detection only**. No base image changes.

> **Review round 2.** Rebased onto `main` after #1037; the CHANGELOG entry sits in the new `[Unreleased]`, not the cut `[0.4.5]`. Both blockers fixed — the gate now builds the Dockerfile GoReleaser publishes with the release toolchain (B1), and the ignore-file wiring is asserted **per trivy invocation** rather than per repository (B2). Also in: the tracking issue is looked up by title across every state (it used to duplicate itself monthly) and stands down when a human reopens it; `image-scan.yaml` gets a `concurrency:` group; the daily scan reads the published Python lines from the schema enum, so `py3.13` is no longer unscanned; `MIN_SELFTESTS` raised to 9; the local-run recipe passes `--build-arg GO_VERSION`. Claim corrections are folded into the text below.

## The policy, and why it is shaped this way

A Debian base permanently carries CVEs that are unfixed in stable and never will be fixed there. A naive "fail on any HIGH" is unsatisfiable on day one, so it gets disabled within a week — and disabled is worse than absent, because it still looks like coverage. Two filters and one split make it survivable.

**1. Fixability before severity.** Severity says how bad a finding is; fixability says whether anyone can act on it today. Only the second works as a gate criterion. On `leoflow-runtime:py3.11`, `--ignore-unfixed` takes **273 findings → 14**, and removes **all five CRITICALs** — every one marked `will_not_fix`, `fix_deferred` or `affected` upstream in bookworm. That is the entire argument in one number.

**2. A severity floor of CRITICAL/HIGH**, applied *after* fixability, for the blocking decision only. Everything fixable is still reported at every severity.

**3. Blocking is split by who can fix the finding**, not by how bad it is:

| Image | Base | When | Blocks? |
|---|---|---|---|
| `leoflow-server` | distroless + our `go.mod` binary | every PR + push | **Yes** — job goes red |
| `leoflow-runtime` (one leg per published Python line) | `python:3.x-slim-bookworm` | daily | No — tracking issue |
| `leoflow-migrate` | `migrate/migrate` (Alpine + vendored Go) | daily | No — tracking issue |

`leoflow-server` has no inherited OS package set: every finding is caused by a commit and closed by a `go get`, so blocking puts the work in front of the person who can do it. **That only holds if the gate scans the artifact we publish**, so it does: it reproduces the GoReleaser recipe — the pinned release toolchain, `CGO_ENABLED=0 -trimpath` — and builds `deploy/Dockerfile.server.release`, the Dockerfile `.goreleaser.yaml` points at. Building `deploy/Dockerfile.server` instead would compile inside a floating `golang:1.x-bookworm` (`go1.27.1` today against the release's `go1.26.6`), and Trivy keys `stdlib` findings to the toolchain in the binary's build info — so it would scan a different vulnerability set, and a stdlib CVE there would be closed by a base rebuild rather than a `go get`, reddening the blocking gate for whoever opened the next unrelated PR. **"Blocks" here means the job goes red**; making it stop a merge needs the check added to the branch-protection rules, which a new job is not by default. The other two inherit package sets we do not author — a CVE lands because a distro security team published an advisory, on a day nobody here pushed anything. Blocking those fails whoever opens the next unrelated PR while the person who can fix it is elsewhere. They get a **daily scan, a SARIF upload, and one self-closing tracking issue** whose body refreshes each run and which comments only when the finding set actually changes (a daily scan that notifies daily is a daily scan nobody reads). That issue is found by title across every state, so a closed one is reopened rather than duplicated, and the scan checks who performed the last reopen before closing — it will not close an issue a maintainer reopened out from under them. The step is `continue-on-error` with an `ERR` trap, so a token without `issues: write` warns instead of reddening a workflow whose header promises it never fails the build.

The daily job builds those images **from the checkout**; it pulls no published tag, and is named for that. A base CVE in an already-published `py3.11` that HEAD no longer produces is invisible to it — `trivy image ghcr.io/…:py3.11` answers that question directly, with no build. Which Python lines it scans comes from the `python_version` enum, the single source of truth since #1031, and a first step fails the run if that enum and the per-line SARIF upload steps disagree.

**Visibility over blocking:** every fixable finding goes to the Security tab, one code-scanning category per image. Unfixed findings are excluded from SARIF too — an alert nobody can ever close is not visibility, it is noise that teaches people to stop opening the tab. The full unfiltered inventory is kept as a 30-day build artifact so no filter here is the only copy of the data.

## Scanner choice: Trivy

The repo already runs `aquasecurity/trivy-action` for the filesystem scan, and `security.yaml` carried a `Container image scanning returns once deploy/Dockerfile.server exists` TODO — that file exists now, and the TODO is retired with this change. Beyond continuity: Trivy resolves Debian's `will_not_fix` / `fix_deferred` status (which is what makes `--ignore-unfixed` trustworthy here rather than a guess), reads OS packages, language packages **and Go binaries** in one pass — the last of which is what surfaced the migrate finding below, and which a pure OS scanner would have missed entirely — emits SARIF natively, and supports a native `expired_at` on ignore entries. Grype is a reasonable scanner but has no equivalent self-expiring ignore mechanic, which is the load-bearing piece of the policy below.

## The ignore list cannot silently become permanent

`.trivyignore.yaml` ships with **zero entries** — the honest state, since the blocking gate is green and the reported images are not suppressed. The mechanism is in place *before* the first exception, which is the only time it can be designed calmly instead of under a red build.

Trivy enforces `expired_at` itself: once the date passes it stops honouring the entry and the finding returns. **Verified against Trivy 0.72** — future-dated suppresses, lapsed does not. That makes the result of neglect *the finding comes back*.

Trivy does not require an entry to have an expiry or a rationale at all, which makes the permanent unexplained entry the easiest one to write. `scripts/check-trivyignore-entries.sh` (new CI gate, with `--self-test`) rejects:

- a missing or placeholder `statement`, or one below a length floor
- a missing, unparseable, already-lapsed, or beyond-180-day `expired_at`
- a top-level section name Trivy would silently ignore (a typo'd `vulnerability:` suppresses nothing while looking configured)
- **and the other end of the chain, per scan** — that **every** step running a trivy scan passes the file, through `--ignorefile`, the action's `trivyignores:` input, or `TRIVY_IGNOREFILE`. Asking only whether *some* workflow still references it is a question a repository keeps answering "yes" to while the gate that matters has quietly gone unfiltered: dropping `trivyignores:` from the blocking gate alone used to give rc=0, because the daily scan still had it. Proving one end does not prove the chain, and proving one link does not prove the others.

Shell lines are read with `\`-continuations joined and comments dropped, and `trivy` counts only in command position — `install /tmp/trivy /usr/local/bin/trivy` installs the scanner, `trivy image --download-db-only` warms a database, `trivy convert` re-renders an already-filtered report, and none of the three is a scan. An unrecognised subcommand **fails closed**: a gate that shrugs at what it cannot parse is the failure mode the file exists to prevent.

A scan that must apply no suppressions declares itself where it runs, with a length-checked reason:

```yaml
# trivy-ignorefile-exempt: this is the unsuppressed copy of record — applying
# .trivyignore.yaml here would make the accepted findings unrecoverable.
- name: Full unfiltered inventory
```

Two scans are exempt: that inventory, and the repository `trivy fs` gate, whose findings come from our own module graph rather than a base image nobody here authored.

That wiring check found a real defect in its own first draft: it read any action's free-form `args:` input as an ignore-file path and failed against the real repo. Fixed, and the shape is now a self-test case.

The self-test is **mutation-tested**: nine guards disabled one at a time, **nine go red**. Three of them survived the first pass and were not accepted as redundant — each turned out to be masked by a *different* rule firing (`len(None)` raising, or "no trivy scan anywhere" catching a case the test meant to attribute elsewhere), so the fixtures were rewritten until each case fails for the reason it claims. The nine cover the per-invocation assertion, the exemption reason floor, the YAML end-mark bleed that would let one step's exemption excuse the step above it, fail-closed classification, continuation joining, command position, database-warming flags, `TRIVY_IGNOREFILE`, and "every scan exempt means the file is read by nothing".

`scripts/check-script-selftests.sh` now discovers Python self-tests alongside shell ones, by the same discovery-by-existence rule it already documents, so `scripts/trivy-report.py` is covered by the gate that already covers the `check-*` scripts. It reports **9 self-tests passing**, and the `MIN_SELFTESTS` floor was raised to 9 to match.

## What the scanner says about our images TODAY

Real scans of the published `v0.4.5` images, fixable findings only:

| Image | Total findings | Fixable | **Would block (fixable C/H)** |
|---|---:|---:|---:|
| `leoflow-server:v0.4.5` | 1 | 0 | **0** |
| `leoflow-runtime:py3.10` | 274 | 15 | **2** |
| `leoflow-runtime:py3.11` | 273 | 14 | **2** |
| `leoflow-runtime:py3.12` | 270 | 11 | **0** |
| `leoflow-migrate:v0.4.5` | 87 | 84 | **50** |

**The blocking gate is green on day one.** `leoflow-server` reports nothing under the policy, so it starts clean and any red is a real regression.

**The runtime base is boring, which is the point.** 5 CRITICALs on every leg, all unfixed Debian (`zlib1g` `will_not_fix`, `perl-base` ×3, `libsqlite3-0`). Nothing to do, and the policy correctly says nothing. What *is* actionable is small: `jaraco.context` 5.3.0→6.1.0 and `wheel` 0.45.1→0.46.2 on py3.10/3.11 — both pip packages, both absent from py3.12, so the matrix legs have drifted. Also fixable but below the floor: `pip` 24.0 (5 MEDIUM), `setuptools`, and `libpcre2-8-0` 10.42-1→10.42-1+deb12u1, an OS package that is fixable and would clear on a rebuild — `runtime/Dockerfile` never runs `apt-get upgrade`, so it inherits whatever the base shipped with.

**The unexpected result — and the reason this was worth doing:** `leoflow-migrate` has **50 fixable CRITICAL/HIGH**, and none of it is Debian noise. It is the third-party Go binary inside `migrate/migrate:v4.19.1`, built with Go 1.25.4:

- **CRITICAL** — `pgx/v5` v5.5.4 (×2), `golang.org/x/crypto` v0.45.0, `google.golang.org/grpc` v1.74.2, Go `stdlib` v1.25.4
- **HIGH ×45** — 20 more `stdlib`, 9 `x/crypto`, 5 `x/net`, 4 `grpc`, 3 `otel`, `go-jose`, `x/text`, plus 2 Alpine `musl`

`govulncheck` structurally cannot see any of this — and the precise reason matters, because the loose version of it is wrong. `golang-migrate` **is** in our `go.mod` and govulncheck does see it, resolved against **our** MVS versions and toolchain. What it cannot see is what upstream vendored into their release binary: 48 of the 50 findings sit in `usr/local/bin/migrate`. The Lite path that runs that image is covered here, not there. The image nobody was worried about is the one on fire, and the fix is ours to make — bump the `FROM migrate/migrate:` pin in `deploy/Dockerfile.migrate`, or build the migrator on our own base. **That is out of scope here** (this PR is detection), and it is exactly what the tracking issue is for: the first scheduled run will open it with this table.

So the honest day-one summary: **the gate that blocks is green, and the scan that reports immediately finds a real, fixable, previously-invisible problem.** That is the shape a scanner should have on the day you install it.

## Conventions

- Pinned action SHAs with version comments; reuses the SHAs already in the tree.
- Job-level minimal `permissions`; `security-events: write` only where SARIF is uploaded. `issues: write` is held by `image-scan.yaml`'s single job on **schedule, push and pull_request alike** — `permissions:` takes no expression, so only the steps are guarded by `github.event_name != 'pull_request'`.
- `timeout-minutes` on every job.
- SARIF upload is skipped on `pull_request`, matching the `#437` note in `security.yaml` about PR runs overwriting the main-branch baseline under a shared category.
- Trivy is installed from a pinned GitHub release tarball with retries in the new workflow, mirroring the `gitleaks` step's reasoning (no Docker Hub, no shared-IP rate limit). `scripts/check-service-image-registry.sh` passes.
- No YAML anchors — `actionlint` resolves them but GitHub Actions does not support them; the duplicated `paths:` lists are commented as deliberate.

## Verification

- `actionlint` (with `shellcheck`) clean on every workflow; `shellcheck` clean on the new script.
- All repo `check-*.sh` gates pass; `check-script-selftests.sh` reports 9/9 and its floor was raised to match.
- Scan → `trivy convert` → SARIF → report pipeline dry-run end-to-end against the real published images with the real ignorefile; SARIF validated as 2.1.0.
- The server gate's new build path run locally: the image embeds `go1.26.6` (the release toolchain), and Trivy 0.72 reports **1 total finding, 0 fixable, 0 would-block** on it — the same numbers as the published `v0.4.5` image, so the gate stays green after the change.
- `check-trivyignore-entries.sh`'s per-scan assertion mutation-tested: nine guards disabled one at a time, **nine go red**, including the literal review mutation (drop `trivyignores:` from the blocking gate while the daily scan keeps it), which the previous shape passed.
- The tracking-issue script driven against a stubbed `gh` over all eight paths — create, quiet update, changed-fingerprint comment, reopen-instead-of-duplicate, close, and the human-reopen case where it now stands down.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1118,6 +1118,27 @@ jobs:
         run: bash scripts/check-service-image-registry.sh
 
   # ─────────────────────────────────────────────────────────────────────
+  # Container-scan ignore list. Every finding we accept must say why and carry
+  # an expiry date, and the file holding them must actually be passed to a scan.
+  # Trivy honours `expired_at` — a lapsed entry stops suppressing, so neglect
+  # makes a finding come back rather than stay hidden — but nothing in Trivy
+  # requires an entry to have an expiry or a rationale at all, which makes the
+  # permanent unexplained entry the easiest one to write. The other half is
+  # wiring: a renamed ignore file, or a dropped `--ignorefile` flag, leaves
+  # entries that look accepted and suppress nothing. Both ends are checked.
+  # ─────────────────────────────────────────────────────────────────────
+  trivyignore-entries:
+    name: Container-scan ignore list is reviewable
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+      - name: Install PyYAML
+        run: python3 -m pip install --user pyyaml
+      - name: Check every accepted finding has a reason and an expiry
+        run: bash scripts/check-trivyignore-entries.sh
+
+  # ─────────────────────────────────────────────────────────────────────
   # Script self-tests. cut-release.sh and the check-* gates hold logic nothing
   # else exercises — the tag normalizers, FLAKE_RE (which decides whether a red
   # run is rerun or stops a release), and the gates' ref comparisons. A defect
--- a/.github/workflows/image-scan.yaml
+++ b/.github/workflows/image-scan.yaml
@@ -1,0 +1,289 @@
+# Scheduled container image scan — the REPORTING half of the container policy.
+#
+# ── Why this is not a PR gate ────────────────────────────────────────────────
+#
+# leoflow-runtime is python:3.x-slim-bookworm and leoflow-migrate is
+# migrate/migrate (Alpine + a third-party Go binary). Neither package set is
+# ours. A CVE lands in them because a Debian security team published an
+# advisory, not because anyone here pushed a commit — so a blocking gate on them
+# fails whoever opens the next unrelated pull request, and the person who can
+# actually fix it (by rebasing a base image or bumping a pinned third-party tag)
+# is not in that PR. That gate gets switched off within a week, and switched-off
+# is worse than never-installed because it looks like coverage.
+#
+# So: this workflow never fails the build. It publishes to the Security tab and
+# keeps exactly one tracking issue current. The blocking gate lives in
+# security.yaml and covers leoflow-server, the one image whose every byte comes
+# from our go.mod.
+#
+# ── Why a fixability filter, not just a severity floor ───────────────────────
+#
+# Debian stable always carries CVEs that are unfixed there and never will be
+# fixed there. At the time of writing leoflow-runtime:py3.11 reports 273
+# findings, of which 5 are CRITICAL and every one of those 5 is unfixed upstream
+# (`will_not_fix`, `fix_deferred`, `affected`). `--ignore-unfixed` takes that
+# 273 down to 14 — the ones a human could act on today. Severity says how bad a
+# finding is; fixability says whether anyone can do anything about it. Only the
+# second is a sane gate criterion, and a policy that cannot be satisfied is a
+# policy that gets deleted.
+#
+# Unfixed findings are excluded from the SARIF upload too. A code-scanning alert
+# that can never be closed is not visibility — it is noise that teaches people
+# to stop opening the tab. The complete unfiltered inventory is kept as a build
+# artifact on every run for anyone who needs it.
+
+name: Image scan
+
+on:
+  schedule:
+    # Daily. Base-image CVEs appear on the distro's clock, not ours, so a weekly
+    # cadence means up to seven days between a fix landing upstream and anyone
+    # here knowing. Offset from security.yaml's 06:00 Monday run.
+    - cron: "20 6 * * *"
+  workflow_dispatch:
+  # The two path lists below are duplicated on purpose: GitHub Actions does not
+  # support YAML anchors, so an alias here would not expand. Keep them in sync.
+  push:
+    branches: [main]
+    paths:
+      - "runtime/**"
+      - "deploy/Dockerfile.migrate"
+      - "migrations/**"
+      - "go.mod"
+      - "go.sum"
+      - ".trivyignore.yaml"
+      - "scripts/trivy-report.py"
+      - ".github/workflows/image-scan.yaml"
+  pull_request:
+    branches: [main]
+    # Same inputs. On a PR this reports into the job summary and stays green —
+    # the author sees what their change does to the image without being blocked
+    # by findings that were already there.
+    paths:
+      - "runtime/**"
+      - "deploy/Dockerfile.migrate"
+      - "migrations/**"
+      - "go.mod"
+      - "go.sum"
+      - ".trivyignore.yaml"
+      - "scripts/trivy-report.py"
+      - ".github/workflows/image-scan.yaml"
+
+permissions:
+  contents: read
+
+env:
+  # Pinned: the scanner version decides what is reported, so an unpinned bump
+  # can move the tracking issue overnight for reasons unrelated to our images.
+  TRIVY_VERSION: "0.72.0"
+  # Matches ci.yaml, so the agent baked into the scanned runtime image is built
+  # with the toolchain that ships.
+  GO_VERSION: "1.26.6"
+  GATE_SEVERITY: "CRITICAL,HIGH"
+  ISSUE_TITLE: "Container image CVEs: fixable findings in published images"
+
+jobs:
+  scan:
+    name: Scan published image bases
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    permissions:
+      contents: read
+      security-events: write   # SARIF upload to the Security tab
+      issues: write            # create/update the single tracking issue
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+
+      # Pinned release tarball rather than an action or a Docker Hub pull: the
+      # same reasoning as the gitleaks step in security.yaml. GitHub release
+      # downloads are reliable from Actions and rate-limited by nothing we share.
+      - name: Install Trivy
+        run: |
+          set -euo pipefail
+          url="https://github.com/aquasecurity/trivy/releases/download/v${TRIVY_VERSION}/trivy_${TRIVY_VERSION}_Linux-64bit.tar.gz"
+          ok=0
+          for i in 1 2 3; do
+            if curl -fsSL "$url" -o /tmp/trivy.tgz; then ok=1; break; fi
+            echo "::warning::trivy download attempt ${i} failed; retrying in 5s"; sleep 5
+          done
+          [ "$ok" = 1 ] || { echo "::error::could not download trivy v${TRIVY_VERSION} after 3 attempts"; exit 1; }
+          tar -xzf /tmp/trivy.tgz -C /tmp trivy
+          sudo install -m 0755 /tmp/trivy /usr/local/bin/trivy
+          trivy --version
+
+      # One DB pull for every scan below.
+      - name: Warm the vulnerability database
+        run: trivy image --download-db-only
+
+      - name: Build images
+        run: |
+          set -euo pipefail
+          for py in 3.10 3.11 3.12; do
+            echo "::group::build leoflow-runtime py${py}"
+            docker build -f runtime/Dockerfile \
+              --build-arg "PYTHON_VERSION=${py}" \
+              --build-arg "GO_VERSION=${GO_VERSION}" \
+              -t "leoflow-runtime:py${py}" .
+            echo "::endgroup::"
+          done
+          echo "::group::build leoflow-migrate"
+          docker build -f deploy/Dockerfile.migrate -t leoflow-migrate:ci .
+          echo "::endgroup::"
+
+      # Scan each image once to JSON (fixable findings, every severity), then
+      # derive the SARIF from that report. The file stem names the image in the
+      # summary, so keep it readable.
+      - name: Scan images
+        run: |
+          set -euo pipefail
+          mkdir -p scan-out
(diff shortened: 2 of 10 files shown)


neochaotic:
Review verified the day-one numbers independently (all three reproduce **exactly**: server 1/0/0, runtime py3.11 273/14/2, migrate 87/84/50) and confirmed the gate is not inert. Two blockers, both about the gate pointing at the wrong thing rather than about the policy, which is sound.

## B1 — blocking. The gate scans a binary we never ship

```
GoReleaser publishes:  deploy/Dockerfile.server.release   (.goreleaser.yaml:114,126)  → Go 1.26.6
The gate builds:       deploy/Dockerfile.server           (security.yaml:138)          → FROM golang:1.27-bookworm
```

Reviewer built both and read the embedded toolchain: `go1.27.1` in the gate image, `go1.26.6` in the release path. Trivy keys `stdlib` findings to the embedded toolchain — demonstrated by this PR's own data, where migrate's `go1.25.4` binary yields 21 HIGH `stdlib` findings. So the gate is structurally blind to a stdlib CVE affecting 1.26.6 and not 1.27, and red for the inverse. Latent today (both scan identically), but it is not scanning the artifact.

**This also undercuts the fairness argument.** `security.yaml:105-113` and the doc justify blocking on `leoflow-server` because "a finding there was introduced by a commit and is closed by a `go get`". `golang:1.27-bookworm` is a floating tag: a Go stdlib CVE is closed by a base-image rebuild, not `go get`, and it turns the PR-blocking gate red for whoever opens the next unrelated PR — precisely the scenario used to justify **not** gating runtime and migrate.

## B2 — blocking. The wiring check does not check the wiring it advertises

`check-trivyignore-entries.sh:132-136` fails only when **no** workflow anywhere references an ignore file. Reviewer constructed it against the real entrypoint:

```
delete `trivyignores:` from security.yaml:111 (the BLOCKING gate)   → rc=0  "trivyignore ok"
also delete --ignorefile from image-scan.yaml:144                    → rc=1
```

So a dropped ignorefile on the gate that matters is caught only if dropped from every workflow simultaneously. The self-test exercises only the zero-reference case, which is why this survived the ten-guard mutation pass — the mutation set never contained it. Assert **per invocation**: every step that runs trivy carries an ignorefile reference.

## Should fix before merge

**The tracking issue duplicates itself monthly.** `image-scan.yaml:246` uses `gh issue list --state open --limit 100`, newest-first. This repo has 130 open issues and the 100th-newest is from 2026-08-11 — a one-month window. Past that the lookup returns empty, a *second* tracking issue opens, and the first is abandoned open with a frozen body. Use `--search "in:title …"` or a label, and `--state all` so a closed one is reopened rather than re-created.

**It re-closes a human-reopened issue, daily.** When BLOCKING is 0 it comments and closes any open issue with that title, with no check of who reopened it. A maintainer reopening it to track the migrate bump gets a comment plus a close every day. Converse gap: `--state open` means a human-closed issue is never reopened when findings return. And on a token without issue-write, `set -euo pipefail` with no `continue-on-error` turns the whole daily job red — contradicting the file's own header, "this workflow never fails the build".

**Cost and race.** `server-image` inherits the trigger block with no `paths:` filter, so every docs-only PR now compiles the control plane in Docker (20-min job), duplicating `helm-ci.yaml:154`. And `image-scan.yaml` has no `concurrency:` group — a push racing the 06:20 cron gives two runs that both see no open issue and both create one.

## Claim corrections

- PR says `issues: write` is "only on the scheduled job". There is one job, and it holds `issues: write` on schedule, push **and** pull_request; only the steps are guarded.
- Job named "Scan published image bases" and the PR says "published v0.4.5 images" — it builds from the checkout and never pulls a published tag. A base CVE in an already-published `py3.11` that HEAD no longer produces is invisible.
- `MIN_SELFTESTS:-5` was not raised while `found` went 6→7.
- The local-run doc omits `--build-arg GO_VERSION`, so a local run builds with `runtime/Dockerfile:15`'s 1.26.3 while CI uses 1.26.6 — the same drift as B1, in text that says "as CI does".
- "Blocks every PR" needs a branch-protection update; a new job is not required by default.

## Verified, do not re-litigate

All three day-one counts reproduce exactly. `--ignore-unfixed` is applied consistently across gate, SARIF and the daily JSON (SARIF: 84 results, 50 at `level: error`). `expired_at` tested against Trivy 0.72 directly — all four behaviours hold, including that a **typo'd section name suppresses nothing**. The pinned action SHA is the version that handles a `.yaml` ignorefile correctly; an older one would have silently voided every entry. `.trivyignore.yaml` does not leak into the existing `trivy fs` gate. Both new entrypoints forward arguments and the no-arg CI path drives the same `check()` the self-test drives.

One precision on the migrate finding, worth a clause in the doc so nobody misreads it later: `golang-migrate` **is** in our `go.mod` and govulncheck does see it — resolved against **our** MVS versions and toolchain. What it cannot see is the versions upstream vendored into their release binary. 48 of the 50 findings sit in `usr/local/bin/migrate`.

PR #1037, opened 2026-09-09
feat(runtime): publish a py3.13 task base, deprecate py3.10, and make the published-Python list singular

### LINKED ISSUE #1031: py3.10 base freezes at its 2026-10-31 EOL, and our two supported-Python lists already disagree

neochaotic:
Closes #1031.

Adds a `py3.13` task base, deprecates `py3.10` with a date instead of dropping it, replaces five copies of "which Python lines do we publish" with one enforced list, and puts a calendar on it so the next one surfaces months out.

## What changed

| | before | after |
|---|---|---|
| Published legs | 3.10, 3.11, 3.12 | 3.10 *(deprecated)*, 3.11, 3.12, **3.13** |
| Source of truth | none — 5 copies, 1 enforced | the `python_version` enum in the authoring schema |
| Copies gated | 0 | 5, by `scripts/check-python-runtime-matrix.sh` |
| EOL signal | a person reading a calendar | weekly, from the scheduled security workflow |

## Why `3.13` and not `3.14`

The ceiling is set by the dbt adapters, not by Airflow. Read from PyPI classifiers on each package's current release:

| package | version | max Python |
|---|---|---|
| `dbt-core` | 1.12.4 | 3.14 |
| `dbt-postgres` | 1.11.0 | 3.14 |
| `apache-airflow-task-sdk` | 1.3.1 | 3.14 |
| `dbt-snowflake` | 1.12.0 | **3.13** |
| `dbt-bigquery` | 1.12.0 | **3.13** |
| `dbt-databricks` | 1.12.5 | **3.13** |
| `dbt-duckdb` | 1.11.0 | **3.13** |

A `py3.14` base would give a user an image where `dbt-postgres` installs and `dbt-snowflake` does not, discovered inside *their* build rather than ours.

The Debian suite is untouched — `bookworm` → `trixie` is a separate decision in flight, and mixing them makes both harder to review and to roll back.

## Why `py3.10` gets a warning at compile, and not at boot or build

The issue asked for the argument, so: three places could carry it.

- **`leoflow compile` (chosen).** The author picking `python_version` is the only person who can change it, and compile is the moment they make the choice, with the file open. Stderr, once per compile, non-fatal.
- **Task boot (agent, in the pod).** Reaches the *operator*, who cannot edit somebody else's `leoflow.yaml`, and repeats on every task of every run of every DAG. That is how a real warning becomes noise people filter out.
- **Build time, inside `runtime/Dockerfile`.** Only fires for people who build the base from a source checkout — nobody in the population this deprecation is about, since they pull the published tag.

The warning also fires for an explicit `base_image` pinning one of *our* published `py3.10` tags (both shapes: `py3.10` and `py3.10-v0.4.5`), because a pin is precisely what keeps someone on a base after we stop rebuilding it. A `base_image` pointing anywhere else stays silent — `python_version` is unused for the `FROM` then, and nagging would train people to ignore the message.

Nothing is dropped in this PR. The `py3.10` leg keeps being built and pushed on every release until 2026-10-31; the removal date is recorded in the schema, announced in the CHANGELOG, and asserted weekly.

## Making the list singular

The schema enum was already the only *enforced* statement — it is embedded in the CLI and `LeoflowConfig.Validate()` rejects everything else. It is now the source of truth in fact too: `domain.SupportedPythonVersions()` / `DeprecatedPythonVersion()` read it, so Go never restates it, and the deprecation metadata lives beside the enum as a `x-leoflow-python-deprecations` vendor keyword rather than in a comment.

`scripts/check-python-runtime-matrix.sh` fails when any copy disagrees:

1. `.github/workflows/release.yaml` — the `runtime-image` matrix that actually pushes
2. `Makefile` — `make runtime-images`
3. `runtime/Dockerfile` — the `PYTHON_VERSION selects the interpreter (…)` comment
4. `runtime/python/pyproject.toml` — `requires-python` floor must equal the *oldest* published leg (the helper is pip-installed **into** the base image: a floor above it makes that leg unbuildable, one below advertises an interpreter we ship no image for)
5. `website/content/reference/configuration.md` — the reference table and the new support table

Following `check-lite-prepull-matches-compose.sh`, including both lessons its header records: the workflow is parsed as YAML rather than grepped, so a commented-out matrix cannot keep the gate green and an unquoted `3.10` (which YAML reads as the float `3.1`) is caught; and a missing file, renamed job or unrecognized declaration is a **FAIL**, never a silent pass.

## The calendar

`scripts/python-eol-warn.sh`, weekly, from the existing `security.yaml` schedule. It is deliberately **not** named `check-*.sh` — `cut-release.sh`'s `run_gates()` globs that prefix, and a release must never be blocked because a date is approaching. It exits 0 for every finding and non-zero only when it cannot do its own job (feed unreachable, feed reshaped, enum gone); a PR run is `continue-on-error` so a third-party outage cannot block a merge.

Findings reach a person as a labelled, deduplicated issue rather than an annotation on a scheduled run nobody opens. Two are worth one:

- a line we still call **supported** within 270 days of upstream EOL — the finding that would have raised #1031 in March instead of September;
- a line we already deprecated still being published past its own `remove_after` — the deprecation that quietly never happened.

An already-deprecated line inside its window is a notice, not an alert. Live run against the real feed today:

```
::notice::python 3.10 — deprecated, EOL 2026-10-31, published until 2026-10-31.
No published Python line needs a decision.
```

## `detect.go` is a different list, not a disagreeing one

The issue reads `internal/setup/detect.go`'s `pythonCandidates` (3.11, 3.12, 3.13, plus a forward-looking 3.14–3.18) as the same list drifted. It is not. That list answers *which interpreter on this host can parse a `dag.py`*; the schema enum answers *which interpreter runs inside the DAG image, on a machine that is not this one*. They overlap only because both are Python — `py3.10` being published while the detector starts at 3.11 is not a contradiction. Both `pythonCandidates` and `ci.yaml`'s `python-parser` matrix now say so in a comment, instead of implying a relationship a reader would try to reconcile.

## Fifth and sixth copies found

- **`docs/api/leoflow-yaml-schema.json`** — the repo-designated canonical mirror of the embedded schema. Already byte-equality-enforced by `TestEmbeddedSchemasMatchDocs` (which is how it was found: the suite went red), so it needed updating but no new gate.
- **`website/content/project/adrs/0003-dag-as-image.md`** — lists three base images under a repository name (`leoflow/python-runtime`) that no longer exists. Given a dated pointer to the live list rather than an edit: a decision record states what was decided, not what is true today.

Also stale and fixed: `internal/cli/workspace.go`'s comment claimed the schema-allowed set was already `3.10|3.11|3.12|3.13`, which it was not.

`articles/2026-06-leoflow-v0-0-2-deploy-devto.md` names the three published tags too, but it is a dated published article; left alone on purpose.

## Verified, not assumed

- `python:3.13-slim-bookworm` **exists and is current** — last rebuilt 2026-09-01, Python 3.13.15.
- The full `runtime/Dockerfile` **builds** at `PYTHON_VERSION=3.13`; the resulting image imports `leoflow_runtime` and `airflow.sdk`, and `leoflow-agent` runs as its entrypoint (`leoflow dev (commit none, built unknown, go1.26.3)`).
- `apache-airflow-task-sdk==1.3.1` installs and imports on 3.13.
- Every dbt ceiling in the table above, from PyPI.
- Python 3.10's EOL date from endoflife.date matches the date recorded in the schema — and the watchdog now re-asserts that agreement weekly.
- **The gate can fail.** Each of the five real copies was broken in the working tree in turn, confirmed red with a message naming the file, and restored. Disabling the gate's single comparison line turns 6 of its 18 self-test cases red; disabling the watchdog's threshold turns 3 of its 13 red. Both restored and green.

## Gates run locally

`go test ./...` green · `golangci-lint run ./...` 0 issues · `actionlint` clean (all workflows) · all 9 `scripts/check-*.sh` pass · `check-script-selftests.sh` → 7/7 · `sync-example-dockerfiles.sh --check` clean.

`MIN_GATES` 8→9 and `MIN_SELFTESTS` 5→7 for the two added scripts — they are `-ge` floors, so leaving them stale breaks nothing today, it just lets the invariant they protect tolerate the loss of one (as happened in #1018).

---

### D1 (blocking) — the warning gave a no-op instruction to the population it exists for

One message was rendered regardless of which field produced the deprecated
version, so `python_version: "3.12"` + `base_image: …:py3.10-v0.4.5` printed
`warning: python_version 3.10 is deprecated` (a version their file does not
contain) and `Fix: set python_version: "3.11" … and rebuild` — inert, because
`resolveBaseImage` returns `cfg.BaseImage` verbatim. Following it and rebuilding
reproduced the identical warning.

Headline and remedy now both branch on the source field:

```
warning: base_image ghcr.io/neochaotic/leoflow-runtime:py3.10-v0.4.5 pins
  Python 3.10, which is deprecated; ghcr.io/neochaotic/leoflow-runtime:py3.10
  stops being published after 2026-10-31.
  Python 3.10 reaches upstream end-of-life on 2026-10-31. …
  Fix: repoint base_image to ghcr.io/neochaotic/leoflow-runtime:py3.11-v0.4.5
  in leoflow.yaml and rebuild. Existing images keep running.
```

The tag suffix is preserved (`py3.10` → `py3.11`, `py3.10-v0.4.5` →
`py3.11-v0.4.5`, `…-rc.1` likewise), so the remedy is the literal string that
replaces theirs. `python_version` is not mentioned on that path, because
changing it does nothing.

Every assertion now reads the **rendered text** on both paths, and each path
asserts the *absence* of the other's message. The old test built the config with
`PythonVersion` empty and asserted only that a warning fired.

### D2 — the gate reconciles status and dates, not just the list

For each `x-leoflow-python-deprecations` entry the gate now asserts the docs
status row says Deprecated and carries the matching `eol`/`remove_after`, that a
non-deprecated row says neither (and declares no removal date), that the prose
paragraph exists and repeats both dates plus the replacement, and that
`release.yaml`'s hardcoded comment names the leg as deprecated with its EOL
date. Columns are located by header name, so a reordered table fails loudly.

The docs `Published until` cell is now the literal `remove_after` date instead
of prose, so it can be reconciled at all.

The self-test fixture deliberately gives 3.10 an `eol` (2026-10-31) **different**
from its `remove_after` (2026-12-31) — in the real repo they coincide, which
would let a gate comparing the wrong column pass every case for the wrong
reason. 12 new cases, **30 total**; disabling the single `deprecations` lookup
turns 11 of them red.

The review's exact scenario now fails:

```
$ # move the deprecation from 3.10 to 3.11 in the schema
FAIL: the published-Python list has drifted from the schema enum.
  website/content/reference/configuration.md
    status row for 3.11: 'Supported (default)'
    schema:  deprecated (eol 2026-10-31, remove_after 2026-10-31)
    This table is where a user learns a line is going away.
  …
```

### D3 — the stale repository name

`base_image`'s description in both schema mirrors now names the real default,
says that setting it makes `python_version` unused for the `FROM`, and says that
pinning freezes the interpreter and OS packages. ADR 0014 gets the same dated
pointer ADR 0003 already had.

### D4 — `leoflow validate` warns

`validate` open-coded `cfg.Validate()` + `errDbtBlockWithDagSource` instead of
calling `checkProjectPreconditions`, so the one command an author runs in a loop
was the one that never warned. It now shares the gate, which also removes the
duplication. Two tests drive the real cobra command.

### Folded in

- **D7** — the warning was **fatal** on a stderr write error. There is no error
  in the signature any more. The test asserts through
  `checkProjectPreconditions`, not the renderer: a renderer returning an error a
  caller may discard cannot demonstrate the property, and the first version of
  this test passed against the mutation for exactly that reason.
- **D6** — digest pins. `strings.Cut(ref, ":")` read `…:py3.10-v0.4.5@sha256:…`
  as the tag `py3.10-v0.4.5@sha256` and matched nothing. `splitImageRef` strips
  the digest first and takes the tag at the last `:` after the last `/`, which
  also stops a registry port being read as a tag. The remedy tells a digest
  pinner to re-pin, since carrying the old digest under a new tag would pull the
  deprecated manifest straight back in. A digest-**only** pin stays silent on
  purpose: it names a manifest, not a line, and resolving it needs a registry
  round-trip in a command that must work offline.
- **D8** — width. Body text was folded at 78 and *then* indented by two (91
  columns), and the headline was not folded at all (189 on the `base_image`
  path). Folding now happens at `78-len(indent)`. A test walks the raw output of
  both paths and fails any over-long line that is not a single unbreakable token
  — an image reference must stay copy-pasteable, so it is allowed to overhang.
- **D5** — on `--build` the warning is re-emitted as one wrapped line after the
  image is built, on both the dag.py and dbt paths.

### Mutations verified

| mutation | caught by |
|---|---|
| headline stops branching on the source field | `…PinnedBaseImageNamesBaseImage` (×3), `…DigestPinnedBase` |
| remedy stops branching (everyone gets `set python_version`) | `…PinnedBaseImageNamesBaseImage` (×3), `…DigestPinnedBase` |
| digest stripping removed from `splitImageRef` | `TestSplitImageRef`, `…DigestPinnedBase` |
| revert to `strings.Cut(ref, ":")` | `TestSplitImageRef`, `…DigestPinnedBase` |
| fold at the declared width, then indent | `…RespectsItsDeclaredWidth` (both paths) |
| warning returns the io error **and** the gate propagates it | `TestDeprecationWarningIsNotFatalOnAFailingWriter` |
| `validate` back to open-coding `cfg.Validate()` | `TestValidateWarnsAboutADeprecatedPythonLine` |
| post-build reminder disabled | `TestRemindDeprecatedPythonAfterBuild` |
| deprecation moved 3.10 → 3.11 in the real schema | gate: 6 findings across the docs table, prose and `release.yaml` |
| docs `Published until` drifts from `remove_after` | gate, naming both dates |
| `release.yaml`'s hardcoded EOL date drifts | gate, naming the date |
| the status/date arm disabled in the gate | 11 of 30 self-test cases |

One mutation did **not** fail and is recorded as a limitation: reverting
`splitImageRef` to the original `strings.Cut` leaves
`TestWarnDeprecatedPythonSilentOnPortedForeignRegistry` green, because a
port-bearing foreign registry ends up silent either way — for the wrong reason
under the old parse. The digest cases cover the parser.

`go test ./...` green · `golangci-lint run ./...` 0 issues · `actionlint` clean ·
all `scripts/check-*.sh` green · `check-script-selftests.sh` 7/7 ·
`check-python-runtime-matrix.sh --self-test` 30/30.

## Cross-PR — noted, deliberately not fixed here

**D9 / #1038.** The deprecation `reason` in the schema names
`python:3.10-slim-bookworm` and is printed **verbatim to users** (CLI warning)
and rendered into `configuration.md`. #1038 moves the suite to trixie and
nothing gates that prose. **Whoever merges second updates the `reason` string**
in `internal/domain/schemas/leoflow-yaml-schema.json` and its `docs/api/`
byte-mirror; the docs paragraph in
`website/content/reference/configuration.md` quotes the same sentence.

**#1034.** Expect a textual conflict on the `MIN_SELFTESTS` line in
`scripts/check-script-selftests.sh` and in `.github/workflows/security.yaml`.

**#1043 (filed).** After #1027 and #1034 land, `MIN_GATES`/`MIN_SELFTESTS` at
9/7 face actuals of 11/8 — they would tolerate losing two gates and one
self-test, reintroducing the #1018 slack this PR argues against. #1043 proposes
deriving the floor from a committed manifest and failing on any difference,
rather than a hardcoded integer somebody has to remember to bump. Blocked on
#1027 and #1034 to avoid resolving the same conflict twice.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1152,6 +1152,25 @@ jobs:
         run: bash scripts/check-lite-prepull-matches-compose.sh
 
   # ─────────────────────────────────────────────────────────────────────
+  # Which Python lines we publish a task base image for is stated in five
+  # files, and only the schema enum is enforced by anything (the CLI's
+  # Validate() rejects everything else). By #1031 they had already drifted in
+  # both directions. Both directions fail in the user's build, minutes in: a
+  # selectable version with no published leg is a `docker pull` 404, a
+  # published leg outside the enum is a compile that refuses a good image.
+  # ─────────────────────────────────────────────────────────────────────
+  python-runtime-matrix:
+    name: Published Python matrix agrees everywhere
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+      - name: Install PyYAML
+        run: python3 -m pip install --user pyyaml
+      - name: Check the published Python matrix against the schema enum
+        run: bash scripts/check-python-runtime-matrix.sh
+
+  # ─────────────────────────────────────────────────────────────────────
   stale-deploy-path:
     name: No stale legacy deploy path
     runs-on: ubuntu-latest
@@ -1168,6 +1187,11 @@ jobs:
   # detector promises compatibility we never verified, which is exactly
   # the silent-regression risk #220 (F2) flagged. Only 3.11/3.12/3.13 are
   # generally available today; bump as 3.14+ ship and stabilise.
+  #
+  # This is the HOST interpreter that parses a dag.py, not the interpreter
+  # inside the published task base image — a different list on purpose, checked
+  # by a different thing (the base-image list is the schema enum, gated by
+  # scripts/check-python-runtime-matrix.sh). #1031 read the two as one.
   python-parser:
     name: Python parser tests (py${{ matrix.python-version }})
     runs-on: ubuntu-latest
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -156,6 +156,19 @@ jobs:
   # testing Pro without the repo (#318). One leg per supported Python so an author
   # picks the interpreter via `python_version`. Multi-arch + keyless cosign mirror
   # leoflow-server/leoflow-migrate.
+  #
+  # The matrix below is a COPY. The source of truth is the `python_version` enum
+  # in internal/domain/schemas/leoflow-yaml-schema.json — the list the CLI's
+  # Validate() actually enforces — and scripts/check-python-runtime-matrix.sh
+  # fails when the two disagree. Publishing a leg nobody can select, or letting an
+  # author select a leg nobody publishes, both surface as a `docker pull` 404
+  # inside the user's own build, which is the furthest point from the cause (#1031).
+  #
+  # py3.10 is DEPRECATED: Python 3.10 goes EOL 2026-10-31 and docker-library/python
+  # stops rebuilding an EOL line the day after, so the base freezes with whatever
+  # OS CVEs it has. The leg keeps being published until then because `base_image`
+  # pins make a silent drop a broken build for somebody else; the schema carries
+  # the removal date and `leoflow compile` warns the author.
   runtime-image:
     name: Build + push leoflow-runtime base (py${{ matrix.python }})
     needs: release
@@ -164,7 +177,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ["3.10", "3.11", "3.12"]
+        python: ["3.10", "3.11", "3.12", "3.13"]
     permissions:
       contents: read
       packages: write
--- a/.github/workflows/security.yaml
+++ b/.github/workflows/security.yaml
@@ -132,3 +132,61 @@ jobs:
           [ "$ok" = 1 ] || { echo "::error::could not download gitleaks v${GITLEAKS_VERSION} after 3 attempts"; exit 1; }
           tar -xzf /tmp/gitleaks.tgz -C /tmp gitleaks
           /tmp/gitleaks dir . --no-banner --redact --exit-code 1
+
+  # ─────────────────────────────────────────────────────────────────────
+  # Python EOL watch (#1031)
+  # Every other signal in this workflow reacts to a CVE that already exists.
+  # A base image that stops being rebuilt produces the opposite: the CVEs
+  # accumulate and the image never changes, because docker-library/python
+  # stops rebuilding an EOL line the day after it goes EOL. So this one asks
+  # the calendar (endoflife.date) instead, and it runs here because this is
+  # where the schedule already lives.
+  #
+  # It does not block anything on a date: the script exits 0 for every finding
+  # and non-zero only when it cannot do its own job (feed down, feed reshaped).
+  # A finding opens an issue instead, deduplicated by the python-eol label, so
+  # it reaches a person rather than an annotation on a scheduled run nobody
+  # opens. Issue-opening is limited to the schedule — a PR must not file one.
+  # ─────────────────────────────────────────────────────────────────────
+  python-eol:
+    name: Python EOL watch
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    # A PR must not go red because endoflife.date is down. The findings still
+    # appear as annotations on the PR run — a contributor editing the enum sees
+    # them — but only push and schedule treat the watchdog's own failure as red,
+    # where a rerun is the whole remedy and nobody is blocked meanwhile.
+    continue-on-error: ${{ github.event_name == 'pull_request' }}
+    permissions:
+      contents: read
+      issues: write
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+
+      - name: Check every published Python line against endoflife.date
+        id: eol
+        run: |
+          # pipefail explicitly: the Actions default shell does not set it, and
+          # `tee` would otherwise swallow the script's own failure exit.
+          set -euo pipefail
+          bash scripts/python-eol-warn.sh | tee "$RUNNER_TEMP/python-eol.md"
+
+      - name: Open a tracking issue
+        if: github.event_name == 'schedule' && steps.eol.outputs.alert == 'true'
+        env:
+          GH_TOKEN: ${{ github.token }}
+          TITLE: ${{ steps.eol.outputs.title }}
+        run: |
+          set -euo pipefail
+          # The label is both the filing category and the dedupe key, so create
+          # it if it is missing rather than failing the run on its absence.
+          gh label create python-eol --color FBCA04 \
+            --description "A published Python runtime base is approaching or past upstream EOL" \
+            2>/dev/null || true
+          existing="$(gh issue list --state open --label python-eol --json number --jq '.[0].number // empty')"
+          if [ -n "$existing" ]; then
+            echo "python-eol issue #${existing} is already open; commenting instead of filing a duplicate"
+            gh issue comment "$existing" --body-file "$RUNNER_TEMP/python-eol.md"
+            exit 0
+          fi
+          gh issue create --title "$TITLE" --label python-eol --body-file "$RUNNER_TEMP/python-eol.md"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,11 @@ adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   in *your* build — a selectable version with no published leg is a
   `docker pull` 404 minutes into a compile, naming an image you never typed.
 
+  The gate reconciles the deprecation *status and dates* as well as the list, so
+  the support table in the configuration reference — the only place you can read
+  when a line stops being published — cannot say one thing while the CLI warns
+  another.
+
 - **A weekly Python end-of-life watch.** Every other supply-chain signal in the
   repo (Dependabot, Trivy, govulncheck) reacts to a CVE that already exists. A
   base image whose Python line has gone EOL produces the opposite signal: it
@@ -54,15 +59,27 @@ adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   every release until then, and images you have already built keep running. This
   is announced rather than dropped silently because a DAG image is built `FROM`
   the base and many projects pin `base_image`, so our choice becomes yours until
-  you rebuild. `leoflow compile` now prints a warning when your project resolves
-  to a deprecated line — including when you pin one of our published `py3.10`
-  tags through `base_image` — naming the removal date and the replacement. The
-  warning is at compile, not at task boot: the author choosing the interpreter is
-  the only person who can change it, and a per-task-pod warning would reach the
-  operator instead and repeat on every run until people filtered it out.
+  you rebuild. `leoflow validate`, `leoflow compile` and `leoflow deploy` now
+  print a warning when your project resolves to a deprecated line, naming the
+  removal date and the fix. The warning is at authoring time, not at task boot:
+  the author choosing the interpreter is the only person who can change it, and
+  a per-task-pod warning would reach the operator instead and repeat on every
+  run until people filtered it out. With `--build` it is repeated in one line
+  after the image is built, because minutes of builder output otherwise scroll
+  it off the top of the terminal.
+
+  **The warning names the field you actually set, and its fix works on that
+  field.** Setting `python_version: "3.10"` is told to change `python_version`.
+  Pinning one of our published `py3.10` tags through `base_image` — including a
+  digest pin such as `…:py3.10-v0.4.5@sha256:…` — is told to repoint
+  `base_image`, and is given the exact replacement tag with its suffix
+  preserved. The two are separated because `base_image` is used verbatim: an
+  author who pins it and is told to change `python_version` can follow that
+  advice, rebuild, and get the identical warning back.
 
   **What to do:** set `python_version: "3.11"` (or `"3.12"` / `"3.13"`) in
-  `leoflow.yaml` and rebuild.
+  `leoflow.yaml` and rebuild — or, if you pinned `base_image`, repoint it to the
+  matching `py3.11` tag and rebuild.
 
 ## [0.4.5] - 2026-09-09
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,64 @@ adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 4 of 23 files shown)


neochaotic:
Review found one blocker and three worth fixing before merge. The two headline claims hold: **the gate cannot be made inert** (the reviewer broke all five reconciled sources in the real tree and got rc=1 naming the file each time, and the CI entrypoint calls the same `check()` the self-test drives — no entrypoint divergence), and **3.13 is the right ceiling**.

## D1 — blocking. The warning gives a no-op instruction to the population it exists for

`compile_python_deprecation.go:53-59` renders one message regardless of which field produced the version. For `python_version: "3.12"` plus `base_image: …/leoflow-runtime:py3.10-v0.4.5`:

```
warning: python_version 3.10 is deprecated; …:py3.10 stops being published after 2026-10-31.
  Fix: set python_version: "3.11" in leoflow.yaml and rebuild. Existing images keep running.
```

Two faults in one message. It reports `python_version 3.10` when their file says `3.12`, and the remedy is inert — `resolveBaseImage` (`compile_build.go:67-72`) returns `cfg.BaseImage` verbatim, so following the instruction and rebuilding reproduces the identical warning. The author concludes the tool is broken and filters it, which is exactly the outcome the PR's own rationale argues against.

The test cannot catch it: `TestWarnDeprecatedPythonPinnedPublishedBase` builds the config with `PythonVersion` **empty** and asserts only that a warning fires, never that the remedy matches. Branch the remedy on `deprecatedPythonForBuild`'s source — a `base_image` pin needs "repoint `base_image` to `…:py3.11`", and the headline must name `base_image`, not `python_version`.

## D2 — the reconciliation is half done

The gate reconciles the version *list* but not the deprecation *status or dates*. Move the deprecation from 3.10 to 3.11 in the schema and `configuration.md:47-51` keeps saying `3.11 | Supported (default)` / `3.10 | Deprecated` with a green gate — and that table is the only place a user learns the date. The date is also hardcoded in three more places (`release.yaml:167-170`, `configuration.md:48`, `:61-63`), which the new machine-readable field was supposed to eliminate.

## D3 — a stale name in the property this PR edited next to

`schemas/leoflow-yaml-schema.json:42` describes `base_image` as overriding "the default **leoflow/python-runtime** base image" — a repository name this PR's own ADR note declares no longer exists. That is the string a schema-aware editor hovers to the author, on the field D1's corrected warning would tell them to change. `adrs/0014-supply-chain-security.md:86` has it too and got no dated pointer, unlike 0003.

## D4 — `leoflow validate` never warns

`validate.go:27` calls `cfg.Validate()` directly and bypasses `checkProjectPreconditions`. `validate` is the sub-second command an author runs iteratively with `leoflow.yaml` open — the exact moment the PR's rationale identifies. `deploy` is covered via `runCompile`; `validate` is the gap. One line.

## Non-blocking, worth folding in

`D7`: the warning is **fatal** on a stderr write error (`compile_python_deprecation.go:60-61` returns the io error, `compile.go:91` propagates), aborting a compile that would otherwise succeed — the PR calls it non-fatal. Should be `_, _ = io.WriteString(...)`. `D6`: digest pins (`@sha256:…`) slip through the `strings.Cut` on `:`. `D5`: on `--build` the warning is buried under minutes of docker output. `D8`: the `Fix:` line is not wrapped and renders at 91 columns against the declared 78.

## Cross-PR, for whoever merges second

`D9`: the deprecation `reason` names `python:3.10-slim-bookworm` and is printed **verbatim to users**; #1038 moves the suite to trixie. Nothing gates that prose. Confirmed all four `-slim-trixie` tags exist, so it is staleness, not breakage.

`D10`/floors: after #1027 and #1034 land, `MIN_GATES`/`MIN_SELFTESTS` at 9/7 face actuals of 11/8 — they would tolerate losing two gates, reintroducing the #1018 slack this PR argues against. Also expect a textual conflict with #1034 on the `MIN_SELFTESTS` line and in `security.yaml`.

## One thing the review found that strengthens the PR

All four capped adapters declare `requires-python = ">=3.10.0"` with **no upper bound**, so pip installs them on 3.14 regardless — the classifier is advisory. 3.14 would be an "installs, then breaks at runtime" hazard rather than a clean resolver error. Holding at 3.13 is better-founded than the PR argues.

PR #1038, opened 2026-09-09
build: move the DAG task base image to Debian 13 (trixie)

neochaotic:
## What & why

Moves the **DAG task base image** from Debian 12 (bookworm) to Debian 13 (trixie) — `runtime/Dockerfile` stage 2.

The reason is **OpenSSL**, not the Debian release calendar on its own.

`python:3.x-slim` links CPython's `ssl` module against the **system** OpenSSL, so that library is where every TLS call made from inside a task terminates: every provider hitting an API, every `requests` call, every warehouse driver. bookworm ships OpenSSL 3.0.x, whose **upstream support ended 2026-09-07**. That is the load-bearing date — Debian LTS will keep backporting into 3.0 for a while, but the flow of upstream fixes it backports *from* has stopped, so the gap between a CVE existing and a patch reaching a task pod widens from here. trixie ships 3.5.x, supported to **2030-04-08**.

The suite is on the same trajectory: Debian 12 left regular security support **2026-06-11** and is oldstable under the LTS team's narrower, best-effort scope; Debian 13 has full Security Team support to **2028-08-09**, LTS to **2030-06-30**.

Upstream has already moved — `python:3.11-slim` and `python:3.11-slim-trixie` resolve to the identical index digest, children annotated `org.opencontainers.image.base.name: debian:trixie-slim`. So `-slim-bookworm` is no longer "the default", it is an explicit opt-in to the old suite. The new line pins `-slim-trixie` rather than dropping the suffix, so a future base-image move stays a reviewable diff instead of a silent consequence of upstream retagging — and that pin is now **enforced**, see below.

---

# Review round 2

Rebased onto `main` (picks up #1037 py3.13/deprecation-gate and #1034 image scanning). The CHANGELOG conflict is resolved into the **new** `[Unreleased]`, with `### Changed` placed ahead of `### Deprecated` per Keep-a-Changelog ordering. Verified the entry did not land in the cut `[0.4.5]` section.

## B1 — the false sentence, and a gate for it

#1037 records the py3.10 deprecation in the authoring schema, and the CLI prints its `reason` **verbatim** on compile, validate and deploy. That reason named a full tag: `python:3.10-slim-bookworm`. After this move it describes an image nobody builds — and the reader it is aimed at is exactly the one who will check, find trixie in the generated Dockerfile, and file the whole warning under stale tooling.

**Not fixed by `s/bookworm/trixie/`.** That is right today and wrong again on the next move, silently. The suite is **dropped**: `python:3.10-slim` is literally correct on every suite, since that tag has always resolved to the current one. Same edit applied to the configuration reference's prose, the CHANGELOG's #1037 and #1034 entries, the image-scanning page and `image-scan.yaml`'s header, so the release notes do not name two different suites for one base.

**New arm 8 in `scripts/check-python-runtime-matrix.sh`.** Reads the suite off `runtime/Dockerfile`'s stage-2 `FROM` and refuses any `python:3.x-slim-<suite>` literal in the schema or the configuration reference naming a different one. A suffix-free reference always passes, and the failure message recommends dropping the suite rather than swapping it. It also refuses a stage-2 `FROM` with **no** suite, which turns the Dockerfile's own argument for the explicit `-trixie` into something enforced.

Scope is deliberate: `docs/api/`'s mirror needs no arm (`TestEmbeddedSchemasMatchDocs` binds it byte-for-byte to the copy this gate reads); `CHANGELOG.md` is excluded, because a released section naming the suite of its own era is a record, not drift. The self-test fixture grew a stage-1 `golang:*-bookworm` line so the cases prove something — an arm that grepped for a suite name would read the builder's.

## B2 — CI for the headline surface

`grep -rn system_packages test/ .github/ examples/` returned nothing. The k3d operator E2E now carries a leg that ships **no Dockerfile**, which is load-bearing: `ensureDockerfile` honours a file on disk, so `system_packages:` bolted onto any existing fixture would have been silently ignored and the leg would have passed while testing nothing (the same shape that hid four defects in one dbt fixture). `base_image` points at the base built moments earlier in the same run, not the published one.

The same run asserts the base image's own properties, each previously guarded by a comment and nothing else: the **Debian suite** (read out of `runtime/Dockerfile`, not hardcoded — a literal `"trixie"` would be one more copy of the fact this PR is about), the **`65532:65532 0700`** home dir, and the **numeric final `USER`** on both the base and the generated `system_packages` image.

**py3.13 on trixie, built.** It had never been built by anyone; the first build would have been in `runtime-image` (`needs: release`), after the tag. Built locally on this change: Debian 13, `openssl`/`libssl3t64` `3.5.7-1~deb13u2`, `apt 3.0.3`, `passwd 1:4.17.4-2`, Python 3.13.15, `/home/leoflow` = `65532:65532 700`, final `USER 65532:65532`, `import leoflow, leoflow_runtime` and `from airflow.sdk import DAG, task` both fine, bookworm-built agent runs in it.

## The TLS regression class — corrected by measurement

The review expected OpenSSL 3.5 to raise the certificate floor and bite a user with a legacy vendor certificate. **Measured, it does not.** Four probes against a local TLS server, bookworm 3.0.20 vs trixie 3.5.7, identical on all four: RSA-1024 leaf refused on both, RSA-1024 CA refused on both, SHA-1-signed leaf refused on both, TLS 1.1 refused on both. Both suites already run at security level 2. A certificate that worked before this release still works.

**What did change is the first flight's size.** OpenSSL 3.5 enables the hybrid post-quantum group `X25519MLKEM768` by default and sends its key share in the ClientHello: **517 → 1525 bytes** for the same `ssl.create_default_context()`. That is the size at which known-broken middleboxes and TLS-inspecting proxies stop working, and it fails exactly the way the review described — a hang or reset at task runtime in a pod, against one endpoint, after a green build. The CHANGELOG names it and gives the workaround, verified rather than repeated: pointing `OPENSSL_CONF` at a file setting `Groups` to the classical list puts the ClientHello back to 517 bytes.

The in-pod HTTPS assertion is in the E2E regardless. A generic completed handshake is the only thing that covers a class like this, which no assertion could have predicted in advance.

## Should-fix items

- `configuration.md:109` — the second `system_packages` row (Defaults table) updated.
- **Rollback caveat** in the CHANGELOG: a `v0.4.5` `base_image` pin gets bookworm back for py3.10/3.11/3.12, but there is no `py3.13-v0.4.5`, so a 3.13 project must also move to `"3.12"`.
- **Stage-1 comment fixed.** "bookworm is fine because Debian 12 LTS still ships ca-certificates" argued the wrong axis and was doubly wrong: `GO_VERSION` is pinned by patch, so a Debian security update never reaches that image in place and its refresh mechanism is the Go bump; and `crypto/x509` reads the **final** image's trust store, not the builder's. The comment now says both, and names the condition that forces the move (the first `GO_VERSION` bump after upstream stops publishing a `-bookworm` variant).

## Confirmed correct, unchanged

`0700`; the stage-1 split; `cut-release.sh:109`'s FLAKE_RE fixture (matches on `failed to solve`, never on the suite, verbatim record of an observed failure); `cicd-deploy.md:140` (the user's own CI runner).

**#1034 is not reddened.** Its blocking gate covers `leoflow-server` — distroless-debian12 plus a binary from our `go.mod` — untouched here. `leoflow-runtime` changes suite under the **daily, non-blocking** scan, and `.trivyignore.yaml` is empty, so no suppression can go stale against the new package set.

## Verification

`go test ./...` green. `make lint` 0 issues. `actionlint` clean. All `scripts/check-*.sh` green (`check-chart-version-matches-tag.sh` is release-only and needs a tag argument). `shellcheck -S warning` clean. `check-script-selftests.sh`: 9/9, `check-python-runtime-matrix.sh --self-test` 35/35.

**Full `test/e2e/e2e.sh` run locally on k3d, green end to end**, including both new sections. **CI: 49/49 green.**

One rerun, disclosed rather than hidden: `E2E split (k3d two-process api/scheduler)` failed first time on `ErrImagePull` — the k3d import flake `test/e2e/lib.sh` exists to paper over, where `k3d image import` reports success and leaves the node without the image. It is not this change: every other k3d job on the same commit passed on the first attempt, including `E2E operators` (which imports one *more* image than before) and `E2E deploy`, and a trixie image that had actually pulled could not fail as a missing image. Green on rerun, no code touched.

Mutations — each applied and asserted to have applied, red confirmed, restored:

| # | Mutation | Caught by |
|---|---|---|
| M1 | schema `reason` says `python:3.10-slim-bookworm` again | gate arm 8: *"names python:3.10-slim-bookworm / stage 2: FROM …-slim-trixie"* |
| M2 | stage-2 `FROM` drops `-trixie` | gate arm 8: *"pins no Debian suite"* |
| M3 | `chmod 0755 /home/leoflow` | E2E: *"/home/leoflow is '65532:65532 755', want '65532:65532 700'"* |
| M4 | base built on bookworm while the Dockerfile says trixie | E2E: *"base image is Debian 'bookworm' but … pins 'trixie'"* |
| M5 | final `USER leoflow` instead of the numeric pair | E2E: *"final USER is 'leoflow', want numeric '65532:65532'"* |
| M6 | `system_packages:` removed from the leg's `leoflow.yaml` | E2E task: *"system_packages did not put curl on PATH"* |
| M7 | trust store emptied, forcing a TLS verification failure | *found a bug in my own assertion* — see below |

M3/M4/M5 were driven by extracting the **shipped** assertion block out of `e2e.sh` by marker and `eval`ing it against mutant images, with the extraction itself asserted so a mutant that failed to apply cannot look like a pass. M6 was driven through the real CLI (`leoflow compile --build`) and the shipped task body.

**A second finding came out of re-running the E2E rather than reasoning about it.** The leg pushed under the default `git describe` version, so the first local re-run after editing the DAG body died on a 409 against the shared dev database. The dangerous half of that is the case where the body did *not* change: the push is accepted and the run silently exercises the stored version — a pass that proves nothing, which is how the chaos harness spent a stretch of time testing nothing. The leg now stamps a unique `--dag-version` per run, as `deploy-e2e.sh` already does. CI never sees either case (fresh Postgres per run); whoever edits this leg next would have seen both.

**M7 earned its keep.** `urllib.request.urlopen` wraps the TLS error in `URLError` and puts the `ssl` exception in `.reason`, so my `except ssl.SSLCertVerificationError` branch was unreachable and every TLS failure would have been reported as *"network, not TLS"* — a diagnostic that actively misleads, in the one place someone would be reading it under pressure. Fixed to classify off `err.reason`; all three branches re-verified (success, TLS-refused, no-egress) and the E2E re-run clean afterwards.

## Not verified

- The **third-party apt repository** case (`sqv` stricter than `gpgv` about legacy key material). Debian's own repositories are exercised by the new E2E leg; a custom repo added in a user's Dockerfile layer is not, and remains the case to watch. It fails loudly at build time.
- The **ClientHello size** finding is measured on this host's images, not against a real middlebox — no proxy of that kind is available to test against.
- py3.13 is built and inspected locally, **not** exercised by an E2E: every E2E still pins `PY_VERSION=3.11`.

### DIFF
--- a/.github/workflows/image-scan.yaml
+++ b/.github/workflows/image-scan.yaml
@@ -2,7 +2,7 @@
 #
 # ── Why this is not a PR gate ────────────────────────────────────────────────
 #
-# leoflow-runtime is python:3.x-slim-bookworm and leoflow-migrate is
+# leoflow-runtime is python:3.x-slim and leoflow-migrate is
 # migrate/migrate (Alpine + a third-party Go binary). Neither package set is
 # ours. A CVE lands in them because a Debian security team published an
 # advisory, not because anyone here pushed a commit — so a blocking gate on them
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,7 @@ adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 - **Container image vulnerability scanning, with a policy designed to stay
   switched on.** `govulncheck` reads our Go module graph and says nothing about
-  the Debian packages inside `python:3.x-slim-bookworm`, or about a third-party
+  the Debian packages inside `python:3.x-slim`, or about a third-party
   Go binary baked into someone else's image — so OS-level and vendored-binary
   CVEs in the images we publish reached a human before they reached CI. Trivy
   now scans all three (`leoflow-server`, `leoflow-runtime` on every published
@@ -120,9 +120,10 @@ adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   **2030-04-08**. The suite itself is on the same trajectory: Debian 12 left
   regular security support on 2026-06-11 and is now oldstable under the LTS
   team's narrower, best-effort scope, while Debian 13 has full Security Team
-  support to 2028-08-09 and LTS to 2030-06-30. Verified on all three interpreter
-  variants (3.10 / 3.11 / 3.12): `openssl 3.5.7-1~deb13u2`, and the image still
-  ends on numeric UID `65532` with `import leoflow, leoflow_runtime` working.
+  support to 2028-08-09 and LTS to 2030-06-30. Verified on all four interpreter
+  variants (3.10 / 3.11 / 3.12 / 3.13): `openssl 3.5.7-1~deb13u2`, and the image
+  still ends on numeric UID `65532` with `import leoflow, leoflow_runtime`
+  working.
 
   **This changes what your `system_packages:` resolves to.** That key emits
   `apt-get install` into the generated DAG image, and apt now resolves against
@@ -137,6 +138,29 @@ adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   watch, since `sqv` is stricter than `gpgv` about legacy key material. Such a
   repository fails loudly at build time rather than silently, but it can fail.
 
+  **Your outbound TLS handshake changes shape, and this is the one to read if
+  your tasks talk to anything behind a corporate proxy.** OpenSSL 3.5 enables
+  the hybrid post-quantum group `X25519MLKEM768` by default and puts its key
+  share in the first flight, which takes the ClientHello from **517 to 1525
+  bytes** (measured, same `ssl.create_default_context()`, bookworm vs trixie).
+  Middleboxes, TLS-inspecting proxies and a few load balancers are known to
+  mishandle a first flight that no longer fits one segment; the symptom is a
+  handshake that hangs or resets **at task runtime in a pod, after a green
+  build** — against one endpoint, from inside your network, which is the hardest
+  shape of failure to attribute. If you hit it, restore the classical group list
+  by pointing `OPENSSL_CONF` at a file containing
+  `[system_default_sect]` / `Groups = x25519:secp256r1:x448:secp521r1:secp384r1`
+  (verified: that puts the ClientHello back to 517 bytes) and tell us, so the
+  base can carry the setting if it turns out to be common.
+
+  **What did NOT change is the certificate floor**, which is worth stating
+  because it is the thing people expect an OpenSSL major bump to break. Both
+  suites already run at security level 2. Measured on four probes against a
+  local TLS server, bookworm's 3.0.20 and trixie's 3.5.7 behave identically: an
+  RSA-1024 leaf, an RSA-1024 CA and a SHA-1-signed leaf are refused on both, and
+  TLS 1.1 is refused on both. If a legacy vendor certificate worked before this
+  release, it still works.
+
   One silent behavior change is pinned rather than inherited: Debian's `passwd`
   went 1:4.13 → 1:4.17.4 and the `HOME_MODE` default tightened with it, so the
   unchanged `useradd` line produced `0755` on bookworm and `0700` on trixie.
@@ -146,13 +170,31 @@ adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   agent build stage stays on `golang:*-bookworm` deliberately; see the comment in
   `runtime/Dockerfile` for why (nothing from it reaches a task pod).
 
+  **Rolling back is not symmetric across Python lines.** Pinning `base_image`
+  to a `v0.4.5` tag gets you the bookworm base again — for `py3.10`, `py3.11`
+  and `py3.12`. There is no `py3.13-v0.4.5`: that leg is published for the first
+  time in this release, so it only ever existed on trixie. A project on
+  `python_version: "3.13"` that needs bookworm has to move to `"3.12"` as well
+  as pin the older base.
+
+  **This path now has CI behind it.** Nothing in the repository built a DAG
+  image with `system_packages:` set, so the apt half of this change was
+  certified by a hand-run build and nothing re-certified it after merge. The
+  k3d operator E2E now compiles a project through the **generated** Dockerfile
+  (no hand-written one, which would have skipped the apt layer entirely),
+  asserts the package landed, and makes one real outbound HTTPS request from
+  inside the task pod. The same run asserts the base image's own properties —
+  the Debian suite (read out of `runtime/Dockerfile` rather than hardcoded), the
+  `65532:65532 0700` home dir, and the numeric final `USER` — each of which was
+  previously guarded by a comment and by nothing else.
+
 ### Deprecated
 
 - **`python_version: "3.10"` — the `py3.10` base image stops being published
   after 2026-10-31.** Python 3.10 reaches upstream end-of-life on that date, and
   `docker-library/python` stops rebuilding an EOL line the day after
   (`python:3.9-slim` was last rebuilt 2025-11-01, one day after 3.9 went EOL).
-  From November, `python:3.10-slim-bookworm` — and therefore
+  From November, `python:3.10-slim` — and therefore
   `ghcr.io/neochaotic/leoflow-runtime:py3.10` — receives no further OS security
   updates and accumulates unfixed CVEs indefinitely.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -109,6 +109,43 @@ adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   documented in [Image
   scanning](https://leoflow.dev/contribute/image-vulnerability-scanning/).
 
+### Changed
+
+- **The task base image moves from Debian 12 (bookworm) to Debian 13 (trixie),
+  which takes its OpenSSL from 3.0.x to 3.5.x.** `python:3.x-slim` links CPython's
+  `ssl` module against the **system** OpenSSL, so this is the library every TLS
+  call made from inside a task terminates in — every provider hitting an API,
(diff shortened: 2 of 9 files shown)


neochaotic:
Review found three blockers. Two are mechanical (rebase), one is a user-visible lie we would ship, and there is a regression class the PR does not mention at all.

## B1 — after the rebase, `leoflow compile` prints a false sentence to py3.10 users

Verified: this branch does **not** contain `b2929f1` (#1037), and the merge conflicts on `CHANGELOG.md`. Once resolved, #1037's deprecation `reason` — printed **verbatim** to users on `compile`, `validate` and `deploy` — reads:

> `… so python:3.10-slim-bookworm — and therefore leoflow-runtime:py3.10 — stops receiving OS security updates …`

A py3.10 author checks their generated Dockerfile, finds trixie, concludes the warning is stale tooling, and ignores it. That destroys the mechanism #1037 shipped to reach them before 2026-10-31.

**The gate is blind to it — measured, not assumed.** The reviewer resolved the conflict and ran `scripts/check-python-runtime-matrix.sh` on the merged tree: `OK`. Its prose arm asserts `eol`, `remove_after` and `replacement`; the suite name is invisible to it. So this PR neither breaks the gate nor is protected by it.

**Do not just s/bookworm/trixie/.** That fixes today and re-breaks on the next suite move. Make the reason suite-agnostic (`python:3.10-slim`, which is also literally correct — that tag already resolves to the trixie child, `sha256:fd76ade0c607f27…`), and add an arm reconciling any `python:3.x-slim-<suite>` literal against `runtime/Dockerfile`'s stage-2 `FROM`. That is #1037's own thesis applied to #1037's own output.

## B2 — the headline surface has zero CI coverage, and the 41 green checks are stale

```
$ grep -rn "system_packages" test/ .github/ examples/
(no matches)
```

Nothing in the repo builds a DAG image with `system_packages:`. The apt 2.6.1 → 3.0.3 / `gpgv` → `sqv` path — the thing the CHANGELOG spends a paragraph on — is certified by a local build only, and after merge nothing re-certifies it.

Compounding: `python-runtime-matrix` (added by #1037) **never ran on this PR**, so all 41 checks are green against a tree two merges stale. And **py3.13 on trixie has never been built by anyone** — #1037 added it to the matrix, every E2E pins `PY_VERSION=3.11`, and the first-ever `python:3.13-slim-trixie` build happens in `runtime-image`, `needs: release`, **after the tag exists**. Release notes would promise `py3.13-v0.4.6` while GHCR 404s.

## B3 — rebase and re-run CI

Mechanical, but it is what makes B2's numbers mean anything.

## The thing nobody mentioned: OpenSSL 3.5 raises the TLS floor

`SECLEVEL=2` — TLS < 1.2 rejected, RSA < 2048 rejected, SHA-1 signatures rejected. A user whose warehouse or vendor API has a legacy cert gets `certificate verify failed` **at task runtime in a pod, after a green build**. That is the PR's own quiet-failure argument, inverted, and it is far more likely to bite than a re-pinned apt version.

Worth knowing: `dbt-e2e.sh:139` genuinely does `pip install dbt-postgres` **on the trixie base** and is green, so the adapter install is certified — better than the PR claims. But `dbt-e2e.sh:80` starts `postgres:16-alpine` with no TLS and `sslmode=prefer` falls back to plaintext, so **zero bytes go through the new libssl**. The property "outbound TLS still works" is evidenced by a version string, not a handshake.

Needs a CHANGELOG sentence naming the TLS floor, and ideally one k3d task doing a real outbound HTTPS request.

## Confirmed correct — do not re-litigate

**0700 is right.** The reviewer checked it against what the executor actually mounts: no `runAsUser` anywhere for task pods, `FSGroup: 65532` only, and every dbt path (`DBT_PROFILES_DIR`/`TARGET_PATH`/`LOG_PATH`) points at the `/tmp` emptyDir, never `$HOME`. pip cache and matplotlib config are owner-writable regardless. Nothing needs group or other.

**The stage-1 split is sound.** Only `/out/leoflow-agent` crosses, `CGO_ENABLED=0`, so `crypto/tls` is pure Go and `crypto/x509` reads the *final* image's trust store. One correction to the reasoning though: "bookworm is fine because Debian 12 LTS still ships ca-certificates" argues the wrong axis — stage 1 is pinned to `golang:1.26.3-bookworm` **by patch**, so its staleness clock is the Go pin, not Debian's calendar.

**Both deliberate non-changes are right**: the `cut-release.sh` FLAKE_RE fixture matches on `failed to solve` independently of the suite, and `cicd-deploy.md:140` is about the user's own CI runner, not the task base.

## Should fix while here

- `configuration.md:109` — a **second** `system_packages` row still says just "apt packages". Only `:31` was updated; the checklist claims the row (singular).
- The chmod ships a comment where this repo ships a gate. Nothing asserts `0700`, nothing asserts stage 2 is trixie, nothing asserts the final `USER 65532:65532`. Two `docker run` lines in an E2E that already builds the base.
- Rollback: a 3.13 project cannot roll back to v0.4.5 (that base was never published). One CHANGELOG sentence.

## Separately, and it undercuts the PR's premise

`security.yaml:71-90` is `scan-type: fs` while the check is **named** "Trivy (image scan)" — green on this PR having scanned no image at all. So "move suites so OS CVEs get patched" currently has no instrument watching either suite. That is #1034's job and it is in review; noting it here because it is the only instrument this PR's security argument could have had.

PR #1046, opened 2026-09-10
fix(api): give readiness its own connection and one budget, and floor the chart timeout

### LINKED ISSUE #1040: readiness: the 2s schema bound is a per-step deadline, not a probe budget
Split out of the review of #1033 (fix for #1023), which asked for these to be filed rather than widened into that PR.

## What is there now

`internal/api/health.go` bounds the schema call at `schemaProbeTimeout = 2 * time.Second`, chosen to sit under the chart's `probes.readiness.timeoutSeconds: 3`. But it is applied **per dependency**, and `Ping` is not bounded at all — it inherits only the request context.

So the handler's total is unbounded in principle: with Postgres and Redis both registered, the worst case is `Ping(pg) + 2s + Ping(redis)`, and each `Ping` can take as long as the kubelet allows. The probe still fails safe (the kubelet times out and counts a failure, which is the same verdict a 503 would produce), so this is a legibility and log-quality problem rather than a correctness one — but a probe that reports nothing produces no log line naming the dependency, which is exactly the diagnostic the #1023 fix exists to provide.

## Shape of the fix

One deadline for the whole handler, derived once at the top and shared by every step, rather than a fresh 2s per dependency. Something like:

```go
ctx, cancel := context.WithTimeout(c.Request.Context(), probeBudget)
defer cancel()
```

with each step inheriting it, so the last dependency in the map gets whatever the earlier ones left rather than a full fresh allowance.

## Related

- The budget is a hardcoded constant with nothing tying it to the operator-settable chart value — see the sibling issue on `values.schema.json`.

### LINKED ISSUE #1041: chart: nothing stops an operator setting probes.readiness.timeoutSeconds below the probe's own 2s bound
Split out of the review of #1033 (fix for #1023).

## The gap

`internal/api/health.go` bounds the readiness schema query at 2s, and the reason it is 2 and not something else is the chart's default `probes.readiness.timeoutSeconds: 3` — the code's bound has to sit *under* the kubelet's, or a slow database produces a probe that reports nothing instead of a probe that reports not-ready.

That relationship is a comment in the Go source. `probes.readiness.timeoutSeconds` is a plain operator-settable value in `helm/leoflow/values.yaml` with no floor, so `--set probes.readiness.timeoutSeconds=1` renders happily and silently inverts it: the kubelet gives up at 1s while the server is still willing to spend 2s answering, and every probe against a database slower than 1s becomes a timeout with no log line naming the dependency.

## Shape of the fix

A `minimum: 3` on `probes.readiness.timeoutSeconds` in `helm/leoflow/values.schema.json`, so the render fails rather than the probe. Worth checking at the same time whether the other probe fields deserve floors (`failureThreshold`, `periodSeconds`) or whether readiness timeout is the only one coupled to a constant in the server.

If #1040 lands first and the budget becomes a single whole-probe deadline, the floor should be expressed against that number rather than the current per-step 2s.

### LINKED ISSUE #1042: readiness: the schema probe competes for the main pool, so saturation fails every replica at once
Split out of the review of #1033 (fix for #1023).

## The failure mode

The readiness probe now takes a **second** pool acquire per probe: `Ping` borrows a connection, and the schema `SELECT` borrows another. Both come from the same `pgxpool` the request path uses, sized by `database.maxOpenConns` (default 20).

Under saturation — the pool fully lent out to in-flight requests — the schema query blocks waiting for a connection, hits the 2s bound, and readiness fails. Every replica shares that condition, because load is what caused it, so they all fail readiness in the same window and the Service can empty. That is a **load**-induced outage dressed up as a database problem, and the failure feeds itself: pods leaving rotation concentrates the same load on the ones left.

It is the mirror image of the dirty-window bug the #1033 review caught — a correct-looking readiness assertion that removes the control plane from rotation for a reason that is not "this pod cannot serve".

## Shape of the fix

A dedicated, tiny (single-connection) health pool, so a probe never queues behind application traffic and never consumes a connection the request path needs. `scheduler.NewLeaderPool` already establishes that pattern in this codebase — a separate pool for a purpose that must not be starved by the main one — so this is following an existing shape rather than inventing one.

Worth deciding at the same time whether `Ping` should move to it as well (today it has the same contention, it just predates the schema check).

## Note on severity

Nothing here is new to #1033 in *kind* — `Ping` already acquired from this pool — but #1033 doubles the per-probe acquires, so it lowers the load at which this bites.

neochaotic:
fix(api): give readiness its own connection and one budget, and floor the chart timeout

Three findings split out of the #1033 review, fixed together because they are
one path and the fixes reference each other's numbers.

**The probe competed with request traffic (#1042).** Readiness takes two pool
acquires — Ping borrows a connection, the schema SELECT borrows another — and
both came from the pgxpool serving the API. Under saturation the probe queues
behind application traffic, hits its deadline and fails. Every replica shares
that condition, because load is what caused it, so they fail readiness in the
same window and the Service can empty: a load-induced outage reported as a
database problem, and one that feeds itself as the pods left in rotation absorb
the load of the ones that left. Both reads now go through a dedicated
one-connection pool, opened by NewPostgres rather than wired by the caller — a
caller that forgets to attach it reintroduces the bug silently, with the checks
map still satisfying HealthChecker and every test green.

**The 2s bound was a per-step allowance, not a budget (#1040).** It applied per
dependency and only to the schema query; Ping was not bounded at all. With
Postgres and Redis registered the handler's worst case was Ping(pg) + 2s +
Ping(redis), each Ping limited only by whatever the kubelet allowed. It still
failed safe — the kubelet times out and counts a failure — but a probe that
reports nothing writes no log line naming the dependency, which is the whole
diagnostic #1023 exists to provide. One deadline is now derived at the top of
the handler and shared by every step, on /readyz and /api/v2/monitor/health
alike.

**Nothing stopped an operator inverting the relationship (#1041).** The server's
bound has to sit under the kubelet's, and that was a comment in Go next to a
floorless chart value: `--set probes.readiness.timeoutSeconds=1` rendered
happily and made the kubelet give up while the server was still willing to
answer. The chart now refuses to render below 3, and says what to lower instead
(periodSeconds, failureThreshold) for operators who want a faster reaction.

The issue proposed a `minimum` in helm/leoflow/values.schema.json. That file
does not exist, and introducing a whole-chart values schema to carry one floor
would newly validate every other value in the chart — a much larger change with
its own rejection risk. A named helper that fails the render is the same
outcome in the shape _helpers.tpl already uses for podDisruptionBudget.enabled
and deployment.strategy.

The floor and the Go constant are two ends of one relationship that neither side
can see, so scripts/check-readiness-probe-budget.sh asserts
probeBudget < floor <= default, with a self-test.

Also clamps MinConns in the single-connection pools. poolConfig sets it from
database.maxIdleConns (chart default 5) and pgxpool does not reject MinConns >
MaxConns — its background health check tries to open the difference and swallows
the puddle.ErrNotAvailable — so NewLeaderPool has been attempting four doomed
connections every health-check period. Both single-conn pools now share one
config helper.

Evidence, all mutation-tested:
- the unit test asserts on the DEADLINE each step received, not on elapsed time;
  a duration assertion passes on the broken code whenever the fakes are fast,
  which is always. Reverting to the per-step shape: "2 health call(s) ran with
  no deadline", "recorded 1 deadlines, want 3".
- the integration test saturates the request pool literally (maxOpenConns 1, the
  one connection held) and asserts the probe answers anyway. Run against a real
  migrated Postgres; reverting to the shared pool: both calls return context
  deadline exceeded after 2.01s.
- helm render, with database.url and redis.url set: default renders at 3;
  --set 1, --set-string 1 (the Argo CD path) and --set 0 all fail with the floor
  message; 3 and 5 render; split api/scheduler mode renders.
- the new gate mutated on the real tree: probeBudget 2s -> 4s fails with the
  remediation naming both files, restores clean.

Closes #1040
Closes #1041
Closes #1042


neochaotic:
Review blockers addressed. The first one was the important one, and it was mine:

**The shared budget broke the property the shared budget was for.** With one deadline across every dependency, the check that fails is whichever happened to be running when time ran out, not the one that spent it. Measured on the review's fixture: 18 of 20 runs blamed a healthy Redis for a slow Postgres. That is the wrong-dependency page #1023 and #1040 exist to prevent, moved one step later — the change was not worth having without the fix. Dependencies are checked in sorted order now, each step is timed, and a `DeadlineExceeded` is reported as a deadline naming the dependency that consumed the budget. Regression test first, failing on the old code in both dimensions (blame and order).

`TestReadinessDistinguishesSchemaFromUnavailable` changed its expected string. Its property is unchanged and still asserted — a slow database must not read as a migration problem, and the answer must still name postgres — but "unavailable" became ambiguous exactly where it mattered.

The other two blockers: `README.md` regenerated with helm-docs, and three helm-unittest cases for the floor. Those close the hole you named precisely — with `_controlplane-deployment.tpl` reverted to the raw value, `check-readiness-probe-budget.sh` still exits 0 while the floor is gone; `helm unittest` fails. Verified in both directions.

Also applied: health pool to 2 connections (the dashboard endpoint is the third caller and any authenticated viewer can drive it); `pgStartupBudget` extended to the health pool's first connect (the boot schema gate reads through it, so the retry loop had a hole exactly where it was written for); the `MinConns` comment corrected — you are right that nothing reaches the network, and the `checkConnsHealth` recycling consequence is the one worth naming; the monitor-budget assertion tightened to the same window form; and the `withProbeBudget` comment narrowed, since the scheduler heartbeat is outside it.

Follow-ups filed rather than folded in: #1048 (leader-pool connection expiry churning leadership hourly — pre-existing, contradicts ADR 0009's own stability claim, and I noted the `MaxConnLifetime = 0` trap), #1049 (no metric for pool exhaustion, whose last symptom this PR removes), #1050 (chart-driven budget, which retires the gate and closes the residual half of #1041 — including the "raise timeoutSeconds and nothing happens" case you found), #1051 (per-replica connection budget in the docs).

Two things I did not do. The gate is still not wired to a per-PR job on `internal/api/health.go`: #1047 adds a `Duplicated version facts agree` job that is exactly the right home, and adding a second job to `ci.yaml` now guarantees a conflict — I will fold it in when that lands. And the EKS validation items are noted for the next RC cluster pass rather than done here.

Issue #1020, opened 2026-09-09
chart: the migration Job fails with a bare driver error when the database does not exist


neochaotic:
Hit while installing on a fresh GKE cluster.

`helm install` fails with:

```
Error: failed pre-install: resource Job/leoflow/leoflow-migrate not ready.
status: Failed, message: Job Failed. failed: 4/1
```

Nothing there says what is wrong. You have to know to go find the Job's pod and read its log, and the log is:

```
error: failed to open database: pq: database "leoflow_test" does not exist
```

Which is the raw driver error. It is accurate and useless: it does not say that `database.url` is a chart value, that the database is expected to exist already, or what to do.

## Why this one is worth fixing

Compare it with two errors the chart gets **right**, both hit in the same session:

```
secretKey must be exactly 32 raw bytes or 64-char hex (got 48 chars).
Generate with: openssl rand -hex 32 (64 hex chars). See ADR 0019.
```

```
leoflow-server: postgres: database schema is not current: schema_migrations is
absent — migrations have not run (apply them: the Helm pre-upgrade migration Job,
or `leoflow db migrate` for Lite); schema v26 is required
```

Both name the cause, the remedy, and where the rule comes from. That is clearly the house standard — the migration path just is not held to it, and it is the **first** thing a new operator hits, before anything else has had a chance to work.

## Suggestion

The migrate Job knows the URL it was handed. On a `does not exist` from the driver, say so in the chart's terms:

> database `leoflow_test` does not exist on `postgres:5432`. Leoflow does not create it — create it (`CREATE DATABASE leoflow_test;`) or point `database.url` at an existing one. See the chart README "datastore compatibility".

Two smaller things in the same area:

- `failed: 4/1` means `backoffLimit` burned four pods on a deterministic error. A missing database is not transient; the Job retries it four times and each retry costs ~20s of install time before the useful error surfaces.
- The NOTES output after a **successful** install is long and ends with the migration paragraph, so the migration image and `make migrate-image` are prominent — while the failure path, which is where an operator actually needs that information, surfaces none of it.

neochaotic:
Scoping from the review of #1054, which found a cleaner mechanism than the one I had ruled out.

[...]

PR #1054, opened 2026-09-10
fix(chart): mount the database CA into the migration Job, and stop it being BestEffort

### LINKED ISSUE #1052: database.caConfigMap is never mounted into the migration Job, so the chart's own managed-Postgres recipe breaks it
Found while reviewing #1045. Pre-existing; #1045 does not cause it and does not fix it. This is the single most likely way the migration Job fails on a real managed Postgres.

## The gap

`_controlplane-deployment.tpl` mounts `database.caConfigMap` at `/etc/leoflow/db-ca` for the server. `helm/leoflow/templates/migration-job.yaml` has **no `volumes` and no `volumeMounts` at all** — yet both consume the *same* `databaseUrl` secret.

`values.yaml` documents the managed-Postgres recipe as a DSN carrying `sslmode=verify-full&sslrootcert=/etc/leoflow/db-ca/ca.crt`. Follow it and you get a cluster where **the control plane comes up and the pre-install migration Job fails**, because that path does not exist in the Job's container.

## Why it does not fail in CI

With no `sslrootcert`, `lib/pq` leaves `RootCAs` nil and falls back to the system trust store — which is why the distroless CA bundle matters at all, and why every kind/k3d test passes: they run a plaintext in-cluster Postgres. With `sslrootcert` pointing at an unmounted path under `verify-full`, it errors out. No CI job exercises that combination.

RDS is the concrete case: its CA is not in the system root store, so `verify-full` against RDS *requires* the mounted bundle. GKE with Cloud SQL's per-instance CA is the same shape.

## Fix

A `volumes`/`volumeMounts` block on `migration-job.yaml` gated on `.Values.database.caConfigMap`, mirroring the deployment's, plus a `helm unittest` case asserting the mount renders when the value is set and is absent when it is not.

## Validation

Not reproducible without a real managed Postgres. Next RC cluster pass: set `database.caConfigMap` to a ConfigMap holding the RDS global bundle, use a DSN with `sslmode=verify-full&sslrootcert=/etc/leoflow/db-ca/ca.crt`, install, and read `kubectl -n <ns> logs job/<rel>-leoflow-migrate -c migrate`.

### LINKED ISSUE #1053: The migration Job has no resource requests, so it is BestEffort and a namespace ResourceQuota rejects the install
Found while reviewing #1045. Pre-existing.

`helm/leoflow/templates/migration-job.yaml` sets no `resources` on the migrate container, so the pod lands in the **BestEffort** QoS class. Two consequences that kind and k3d never show:

**1. A namespace `ResourceQuota` on `requests.cpu` or `requests.memory` rejects the pod at admission** — "must specify requests.cpu". Since this is a `pre-install`/`pre-upgrade` hook, that fails `helm install` outright. Quotas like this are routine in enterprise EKS/GKE namespaces, which is exactly the audience for the Pro chart.

**2. BestEffort is first out under node pressure.** An eviction mid-`up` leaves `schema_migrations.dirty = true`. `backoffLimit: 3` then retries into `Dirty database version N. Fix and force version.` — a hook Job that cannot self-heal, on a cluster where the control plane has not started yet.

## Fix

Add a `migrations.resources` value with a modest default (the work is a handful of DDL statements) and render it. Guaranteed or Burstable both solve the quota case; Burstable with a real request is enough for the eviction case.

Worth deciding at the same time whether the dirty-state recovery deserves a documented runbook step (`migrate force`), independently of the resources fix — `test/release/rc-cluster-validation.md` already touches this area.

## Validation

Apply a namespace quota on `requests.cpu`/`requests.memory` and install; expect rejection at admission. Separately, force-delete the migrate pod mid-`up` (or cordon and drain its node) and check whether `dirty` ends up true and whether the retries recover or wedge.

neochaotic:
fix(chart): mount the database CA into the migration Job, and stop it being BestEffort

Two defects in the same template, both found while reviewing #1045 and both
pre-existing. They share a property worth stating: neither is reachable from any
test this repo runs, because both need a real cluster or a real managed
database to appear at all.

**The CA bundle was mounted in the control plane and not in the Job (#1052).**
Both consume the SAME databaseUrl secret, so the recipe values.yaml documents
for managed Postgres — `sslmode=verify-full&sslrootcert=/etc/leoflow/db-ca/ca.crt`
— pointed the Job at a path that did not exist in its container. Following the
chart's own instructions produced an install where the server came up and the
pre-install hook failed.

It hid because the fallback is silent in the safe direction: with no
`sslrootcert`, lib/pq leaves RootCAs nil and uses the system trust store, and
every Postgres in CI is in-cluster and plaintext. Only a database whose CA is
outside the system roots — RDS, Cloud SQL — reaches the failing path, and
nothing in kind or k3d can be one.

**The Job declared no resources, so it was BestEffort (#1053).** Two costs, both
invisible in CI. A namespace ResourceQuota on requests.cpu/requests.memory
rejects the pod at admission, and because this is a pre-install hook that fails
`helm install` outright — routine in enterprise namespaces, which is the chart's
audience. And BestEffort is first out under node pressure: an eviction mid-`up`
leaves schema_migrations.dirty, and backoffLimit retries into "Dirty database
version N. Fix and force version." on a cluster whose control plane has not
started yet. The defaults are small on purpose (50m/64Mi requests) — the work is
a few dozen DDL statements.

Four helm-unittest cases, both fixes mutation-tested: removing the CA block
turns the #1052 cases red, removing the resources block turns the #1053 cases
red, and each mutation's character delta was checked before reading the exit
code.

One correction while documenting: `--set-json 'migrations.resources={}'` does
NOT clear the defaults. Helm coalesces chart values back into a bare empty map,
so the render is unchanged; explicit nulls are required. The values comment
first claimed `{}` worked — verified, and it does not.

The installation guide's quickstart used `sslmode=verify-full` with no
`sslrootcert` and no mention that most managed offerings sign with a CA the
system store does not carry, which is the same gap one layer up. It now shows
the ConfigMap path.

Closes #1052
Closes #1053


neochaotic:
Review findings applied. The three surviving mutations were the substance, and they were worse than "the test is weak" — two of them reproduce the exact wedge this PR exists to remove, and both survived helm-unittest, `helm lint`, kubeconform **and** kind-e2e:

- a wrong `items[0].key` renders valid YAML, then the kubelet cannot set the volume up. The container never starts, so `backoffLimit` never engages and the pod sits in `ContainerCreating` until Helm's timeout.
- a dangling `volumeMounts[0].name` is rejected by apiserver PodSpec validation, so the hook Job is never created.

Both asserts are now whole-object `contains:` copied from `deployment_test.yaml`, plus the container name. Five mutations, each verified applied before reading the exit code, all red.

**Your highest-value item was right and I took it.** The new code path had zero server-side validation anywhere — neither the kubeconform render nor kind-e2e set `database.caConfigMap`, so nothing ever rendered the block. kind-e2e now mints a throwaway CA ConfigMap and installs with it; the DSN stays `sslmode=disable` because this is not testing TLS, only that the volume is declared correctly. `--reuse-values` carries it into the upgrade, so the pre-upgrade hook gets the same coverage for free.

Resources retuned to 2:1 on both axes (100m/128Mi against 200m/256Mi). Your `maxLimitRequestRatio` point is the kind of thing I would not have found: a common value of 4 would have rejected the pod at admission and reintroduced #1053's own failure by another route. The test now pins the limits rather than asserting they exist, since `exists:` cannot see a limit below its request.

Every docs correction taken, including the one that stung: the `kubectl create configmap` line ran before the namespace existed, in the documentation for a fix about following the chart's own instructions. Cloud SQL's `verify-full` caveat is now called out, Azure is removed from the private-CA list (Flexible Server chains to DigiCert Global Root G2, which is in the system store), the quickstart points forward, and `Dockerfile.migrate`'s comment no longer asserts the opposite of the code. I also folded in the lib/pq-vs-pgx asymmetry and the `sslrootcert=system` trap, since that is now the only surviving way to hit #1052's shape.

Filed rather than folded in: #1055 (the Job's pod labels match the control-plane Service selector, so on **upgrade** — not install — the migrate pod can join the EndpointSlice and black-hole API traffic; includes the NetworkPolicy variant and how to measure it on the RC) and #1056 (the Job ignores `nodeSelector`/`tolerations`/`affinity`/`priorityClassName`, same install-wedge class as #1053).

Not taken: `GOMAXPROCS=1`. You called it free insurance and irrelevant at this workload size, and I would rather not add speculative runtime config to a hook that bricks installs. Your #1020 scoping is better than my reasoning for deferring it — "no shell in distroless" stopped being the binding constraint when #1045 moved the build into our module — and I have put the three-commit shape you described on the issue.

PR #1158, opened 2026-09-17
feat(helm): make OIDC reachable from values, so SSO can be turned on

alissonrosa-lang:
## The problem

The server has had full OIDC since ADR 0057. No Helm deployment can use it.

Two things block it, and either one alone is enough:

1. `helm/leoflow/values.yaml` exposes **no `auth.oidc` key at all**. The `auth:` block has `jwtSecret`, `existingSecret`, `tokenTtlSeconds`, `agentTokenTransport`, `secretLivenessMode`, `secretScoping`, and nothing else.
2. `validateOIDC` (`internal/config/server.go`) requires `auth.oidc.tenant_claims`, which is a **map**. `decodeDottedOIDCMaps` reads it only from the YAML file named by `LEOFLOW_CONFIG`, and the project's own error text says so: *"auth.oidc.tenant_claims is a map, so it loads ONLY from the YAML config file named by LEOFLOW_CONFIG; no LEOFLOW_AUTH_OIDC_* environment variable can carry it."* The chart mounts no such file and has no `extraVolumes` to mount one with.

So SSO is not misconfigured in a Kubernetes install. It is **unreachable**, and the boot fails closed.

## The shape of the fix

- **Scalars as env.** All 12 render as `LEOFLOW_AUTH_OIDC_*`, which is the pattern the chart already uses everywhere else.
- **Only the maps in a file.** `tenant_claims` and `role_mappings` go into a ConfigMap as a deliberately **partial** `config.yaml`, mounted read only. Partial is safe, and the reason is in `LoadServer`: viper reads the file first and the environment after, so every other setting still arrives by env and outranks the file. I verified that end to end, feeding the rendered output of `helm template` into the real `LoadServer` + `Validate`.
- **Client secret in the Secret**, never in the ConfigMap, with its own `auth.oidc.existingSecret`. The JWT signing key and the IdP secret rotate on different clocks and should not share one field.
- **`checksum/oidc-config` annotation**, because `LoadServer` parses the file once at boot: without it, an edited map sits on disk unread until someone restarts the pod by hand.
- **Three render guards** fail the template instead of the pod: missing tenant pin, non-`https` issuer, and an `extraEnv` that hijacks `LEOFLOW_CONFIG`. The server already reports every missing key in one error; this does the same one step earlier, where an Argo CD sync can show it. Argo never renders `NOTES.txt`.

## Nobody who does not use SSO sees a change

With `auth.oidc.enabled: false` the render is **byte-identical** to today, verified by `diff` in three profiles: default, split with s3 logs, and bootstrap with `secretKey`. No existing key changed default or meaning.

## Commits

Two-commit pattern per ADR 0011, plus two.

| commit | what |
|---|---|
| `test(helm)` | the 13 cases, **failing 12 of 13**. The one that passes is the disabled case, and it passes because nothing renders it yet. |
| `feat(helm)` | the implementation. 286 of 286 tests pass. |
| `fix(scripts)` | a latent bug in `helm-template-checks.sh` that the new example fixture surfaced, see below |
| `docs(changelog)` | `[Unreleased]` entry |

## The script bug, which is separate scope

`helm-template-checks.sh` runs under `set -e` with `pipefail`. The loop validating `secretKey` pipes `grep` into `head` and `sed`, so a fixture that legitimately has no `secretKey` makes `grep` exit 1 and **the script dies right there, mid-loop**. Every assertion below it, the entire agent-TLS auto-gen section, stops running, while the output shows nothing but `OK` lines and no error.

The three existing fixtures all have `secretKey`, so it never fired. The OIDC example does not, which is how it surfaced. A check script that quietly stops checking is worse than the bug that revealed it.

Happy to split this into its own PR if you prefer.

## What I validated

- `helm lint` with OIDC on, off, and with the example values
- `helm unittest`: **286 passed, 25 suites**
- `helm-template-checks.sh`: all contracts met, including 12 new OIDC ones and the assertion that the client secret is absent from the ConfigMap
- `helm-docs -c helm/leoflow`: README regenerates identical, so the `helm-ci` gate passes
- The rendered env plus ConfigMap fed into the real `LoadServer` + `Validate`: the scalars survive from env, the dotted map keys survive, `scheduler.loop_interval_ms` stays at its default

## What I did not validate

**No cluster, no real IdP.** There is no end-to-end proof that a Google login completes. What is proven is that the configuration the chart produces is the configuration the server accepts at boot.

## Known sharp edge

Turning this on makes SSO the only login path except for `breakGlassEmails`. An install with an empty `breakGlassEmails` and a wrong `tenantClaims` locks everyone out. I put that in `NOTES.txt` as a warning rather than a render refusal, because refusing would block a deliberate SSO-only deployment. But Argo CD does not render `NOTES.txt`, so in a GitOps install that warning reaches nobody. Worth a second opinion on whether it should be louder.

## Example

`helm/leoflow/examples/values-oidc-google.yaml`, Google Workspace pinned on the `hd` claim, which by itself excludes `@gmail.com`. Entra is the same file with `tenantClaim: tid`, a GUID on the left of `tenantClaims`, and `roleMappings` actually doing something, since Entra emits groups.


neochaotic:
Superseded by #1159, which is merged. Same capability, carried forward with the chart's Secret ownership and `documentIndex` corrected, plus the tenant-pin requirement the chart now refuses to render without. Thanks for the groundwork: the values shape here is what #1159 shipped.

PR #1167, opened 2026-09-17
fix(api): return a refused single sign-on to the login page, not to raw JSON

neochaotic:
The last piece of the SSO journey that a user actually sees.

Every fail-closed path answered `problem+json` with 403. Both OIDC routes are reached **only** by a top-level browser navigation: the user clicks the sign-in control, or the IdP redirects them back. So a denied login rendered as a page of raw JSON, with no way back to the sign-in page, which is where the retry and the break-glass form live.

The browser now returns to the login page, which says that sign-on did not complete and where the reason is recorded.

## What is deliberately not in the redirect

**The reason.** Withholding it from the browser is the posture of this entire path, since it is the same information someone probing the deployment is after. The marker is `sso_error=1` and nothing else. The cause is in the audit row and, since #1162, in a WARN on the server log.

The marker is also ignored unless a flow exists, so it cannot be used to make a JWT-only deployment claim a sign-on was rejected.

## The tests changed shape, and that is the interesting part

A denied login is a **302** now. So `rec.Code == http.StatusFound` no longer means the login worked, and twelve assertions that read it exactly that way would have passed on a rejection. One of them stated the inference out loud:

```go
if rec.Code == http.StatusFound {
    t.Fatal("callback minted a session for a tenant-mismatched returning user; want rejection")
}
```

Both outcomes are now asserted in one shared place each, so they cannot drift back together:

- `assertLoginDenied` - no session minted (the security property; the rest is UX), the redirect stays on this origin (a deny runs before anything is authenticated, so an off-origin one would be an open redirect), and the body carries no reason.
- `assertLoginSucceeded` - a session cookie exists, and the redirect is **not** the bounce back to the login page.

Verified in both directions rather than assumed: reverting the redirect to `AbortProblem` fails eleven tests, and making the success path redirect like a denial fails four.

One site is deliberately neither helper. `startLogin` asserts the redirect **out** to the IdP, where no session exists yet, so it checks only that it is a redirect and that it is not the bounce back. The first draft of this change applied the success helper there too and it failed, correctly.

`golangci-lint`: 0 issues.


neochaotic:
Superseded by #1169: this PR's head got stuck on a stale commit after a rebase and GitHub would not resync it. Same branch, same commits, nothing dropped.

PR #1171, opened 2026-09-17
fix(server): warn when no break-glass address can actually sign in

neochaotic:
Stacked on #1168; GitHub will retarget this to `main` when that merges.

#1168 warns when `break_glass_emails` is **empty**. A non-empty list was treated as safe. It is not, and this is the worse of the two, because the operator believes they have a way back in.

## What happens today

The gate admits the address. Then `FindUserByLogin` finds no row and answers the same `invalid credentials` a wrong password gets.

The only local password user that ever exists is the bootstrap admin, and only when `bootstrap.password` was set at first install (the chart renders `LEOFLOW_BOOTSTRAP_PASSWORD` but never `LEOFLOW_BOOTSTRAP_EMAIL`, and the default is empty). Every other local user comes from `POST /api/v2/users`, which is admin-authenticated: **exactly what cannot be reached during a lock-out.**

So writing your own work address into `breakGlassEmails` and expecting it to work is the natural mistake, and nothing said otherwise.

## Why this is a SQL question and not a Go one

A row is not enough. A user created by OIDC just-in-time provisioning has a **NULL `password_hash`** (the `users_has_auth` constraint permits it, because the OIDC subject is the other half), so the row can exist while no password can ever verify against it. `CountLocalPasswordUser` requires `password_hash IS NOT NULL AND is_active`.

The tenant checked is `default`, not a configured one: the login page sends no tenant and `authTokenHandler` defaults an empty one to `default`, so that is the only tenant a break-glass login can land in.

One working address is enough to get back in, so this warns only when **none** of them works, and then names all of them.

## Verified against a real database, in both directions

Not only against the fake:

```
auth.oidc.break_glass_emails lists "ops@corp.example", "sre@corp.example", and none of
them has a local password login in the "default" tenant, so none of them can sign in.
```

Then inserting a local password account for `ops@corp.example` and rebooting: the warning count goes to zero. Note `Sre@Corp.Example` came back normalized, matching how the login path normalizes it.

The integration test pins the distinction a fake cannot make: a local password user is a way in, an OIDC-provisioned row is not, a missing tenant is absent rather than an error, and case is not significant.

## One test detail worth stating

The fake lowercases and deliberately does **not** trim, mirroring `CountLocalPasswordUser`, which compares `lower(email) = lower($2)` and trims nothing. A fake that normalized more than the query does would hide a caller that forgets to trim, while the login path does trim, so the two would disagree in production with every test green. Removing the trim from the caller now fails a test; I checked.

`golangci-lint`: 0 issues. Every `scripts/check-*.sh` passes.


neochaotic:
Superseded by #1173. This branch carried #1168's commits, which squash-merged into main, so a rebase replayed them as conflicts. Same single commit, cherry-picked onto a clean base, nothing dropped.

PR #1209, opened 2026-09-20
feat(gcp): put the node TTL on before the pool is full, not after

### LINKED ISSUE #1207: Apply the node TTL to a one-node pool, then resize, instead of rolling every node
`provision.sh` creates the pool at its full size and then applies
`--max-run-duration`, which **recreates every node one at a time**. At ten nodes
that took over thirty minutes and outran gcloud's own operation wait, which is
what #1206 fixes the handling of.

#1206 makes that survivable. This would make it not happen.

## The shape

```
clusters create --num-nodes 1
node-pools update --max-run-duration ...   # rolls ONE node
clusters resize --num-nodes 10             # the other nine come up WITH the TTL
```

Cheaper in the obvious way (about 35 minutes less ten-node time per run, roughly
USD 1) and in a less obvious one: the window in which a cluster exists with **no
TTL at all** shrinks from the whole roll to a single node's recreation. That
window is the one this directory's guardrails are weakest in, because the node
TTL is the only safeguard that survives the script being killed.

## The assumption that has to be checked first, on a real cluster

**Does a node added by `clusters resize` inherit `maxRunDuration` from the pool's
node config?** It should: `maxRunDuration` lives under the node pool's
`config`, which is the template new nodes are built from. But "should" is not
good enough here. If later-added nodes do NOT inherit it, this change puts the
guardrail on 1 node out of 10 and removes it from the other 9, which is
categorically worse than the slow roll it replaces.

Checking it is cheap and bounded:

```
create 1 node, apply the TTL, resize to 2,
then read maxRunDuration back from the POOL and confirm BOTH instances
carry the expiry (gcloud compute instances describe, not just the pool config)
```

About fifteen minutes and well under a dollar. The pool config is not sufficient
evidence on its own: the pool template saying `maxRunDuration` and a running
instance actually carrying an expiry are two different claims, and this directory
has already been bitten four times by exactly that gap between what a command
accepts and what it does.

## Do not land this without that check

The read-back `provision.sh` already performs reads the **pool's** config. If
resize-added nodes silently skip the expiry, that read-back would pass while nine
nodes ran unbounded, and the failure would be invisible until a bill. If the
check comes back negative, close this: the slow roll is correct and the cost is
the price of the guarantee.

neochaotic:
Closes #1207.

Applying `--max-run-duration` recreates every node in the pool, one at a time. At
ten nodes that ran **over half an hour**, outran gcloud's own operation wait, and
left a window in which a ten-node cluster existed with **no expiry on it at all**.

That window is where every guardrail here is weakest, because the node TTL is the
only one that survives this script being killed. Everything else depends on
something of ours continuing to run.

So the pool is created with **one** node, the TTL goes on while there is one node
to recreate, and the rest arrive afterwards via `clusters resize`, born with the
expiry already in force.

## The claim this rests on was verified first, on a real cluster

Getting it wrong would put the guardrail on 1 node out of 10 and remove it from
the other 9 — categorically worse than the slow roll it replaces. The pool's own
config does not prove it, so I inspected the **instances**.

`leoflow-exp-provision-probe-09201213`: created with 1 node, TTL 40m applied,
resized to 2, both instances inspected:

| instance | created | `scheduling.maxRunDuration.seconds` |
|---|---|---|
| `gke-...-96hm` | 08:17:36 (before the TTL) | **2400** |
| `gke-...-zmrw` | 08:23:05 (**by the resize**) | **2400** |

Cost of the check: ~USD 0.20. Cluster torn down.

## And it is not taken on faith at runtime either

After the resize the script lists the instances and **deletes the cluster** if any
came up without an expiry. The pool reporting `maxRunDuration` and a running
instance actually carrying one are two different claims, and this directory has
been bitten four times by exactly that gap between what a command accepts and
what it does.

## The saving is not the reason

Measured: about **18 minutes and USD 0.60 a run**, not the 35 minutes I estimated
in #1207 — the roll takes ~20 minutes even for a single node. I corrected that on
the issue when the numbers came in.

The shorter unbounded window is the reason.

## Self-tests

`seed_nodes` is checked at 1, 2, 3 and 10, because its two obvious wrong
spellings are both silent: seeding with `$NODES` restores the roll this exists to
remove, and seeding with `1` unconditionally turns a one-node run into a resize
to itself.

31 self-tests pass.


neochaotic:
Marcada como draft ate ser exercitada contra um cluster de verdade.

O que esta validado aqui e a **premissa**: um no criado por `clusters resize` herda `maxRunDuration`, medido nas instancias da sonda `leoflow-exp-provision-probe-09201213`. O que **nao** esta validado e este codigo: a sequencia seed-1 / TTL / resize / verificar instancias nunca rodou. A corrida de 10 nos em andamento usa o caminho antigo (#1206), nao este.

O CI nao cobre essa diferenca. Sem credencial de nuvem nos runners, os ~55 checks rodam os self-tests puros e passam por todo o trecho que fala com o gcloud sem executa-lo. Uma PR verde aqui significa `seed_nodes` correto para 1, 2, 3 e 10, e nada sobre o resize.

Sai do draft quando uma corrida real tiver provisionado por este caminho, com a evidencia colada aqui.

neochaotic:
Validado contra cluster real, que era a condicao para sair do draft.

Corrida `leoflow-exp-pod-per-task-09201532`, 10 x e2-standard-4, us-central1-a.

O cluster nasceu com **1 no**, recebeu o TTL, e so entao cresceu:

```
==> scaling default-pool from 1 to 10, with the TTL already in force
```

Depois disso a corrida seguiu inteira: seis niveis, 1274 observacoes, veredito escrito, teardown verificado. A verificacao das instancias pos-resize nao disparou, ou seja todos os dez nos subiram carregando a expiracao.

**O ganho medido.** O roll do TTL sobre um no acabou em poucos minutos contra os ~40 minutos que ele levava sobre dez, e a janela em que existe cluster **sem expiracao nenhuma** encolheu de dez nos para um. Essa janela e o que importa: o TTL do no e a unica protecao que sobrevive ao script ser morto.

Sobre a estimativa de tempo eu errei tres vezes nesta issue e na #1207 (35 min, depois 18, depois 75). O numero medido aqui e o unico que vale, e ele vem de uma corrida e nao de extrapolacao.

